PR #5, opened 2024-04-12
[ECO-1556] Enforce valid emojis in a Move contract

xbtmatt:
# Description

- [x] Parse all valid utf-8 representations of unique individual emojis that can fit within 10 bytes, hence can function as a symbol for an Aptos coin.
- [x] Revisit the decision to include unquailfied and/or minimally qualified emojis. Fully qualified might be the way to go so we don't run into issues with displays not showing correct things. (Specifically with the variant selector #16: \uFE0F)
- [x] Only include fully-qualified, to avoid potential rendering issues on platforms that do not render the optional non-fully-qualified format
- [x] Verify package publishability

# Checklist

- [x] ~Did you update relevant documentation?~
- [x] Did you add tests to cover new code or a fixed issue?
- [x] ~Did you update the changelog?~
- [x] Did you check off all checkboxes from the linked Linear task?

PR #6, opened 2024-04-14
[ECO-1562] Begin preliminary struct architecture

alnoki:
# Description

1. Escape emoji words globally since they will likely occur in other contexts
2. Refactor package layout for `emojicoin_dot_fun::emojicoin_dot_fun` structure
3. Escape a flake warning
4. Add a stub to the supported emoji set to enable https://github.com/econia-labs/emojicoin-dot-fun/pull/5#discussion_r1564425037
5. Add struct layout:

The registry is stubbed to enable https://github.com/econia-labs/emojicoin-dot-fun/pull/5#discussion_r1564425037.

Markets are also stubbed as an object, to enable coin factory bytecode operations.

Struct layout:

`RegistryAddress` is stored under `@emojicoin_dot_fun` as a means for indexing: a view function can allow an offchain element to determine the address of the registry object by checking the `RegistryAddress`.

The `Registry` is an object complete with an `extend_ref`, which allows it to publish markets at deterministic object addresses. New markets may be published via `object::create_named_object` for the `Registry.extend_ref` as the signer, and the emoji bytes of the market as a seed. This will allow for deterministic passing of type arguments during market init calls via public entry functions, since the expected object address can be determined offchain, then packed into an SDK type argument call that will correspond to types published from coin factory bytecode stored onchain.

That is, the `Emojicoin` and `EmojicoinLP` types will be published under the `Market.extend_ref` signer, which is a function of the `Registry.extend_ref` address and the `emoji_bytes` for the market

# Testing
From `src/move`:

```sh
git ls-files | entr -c aptos move compile --dev
```

# Checklist


src/move/sources/emojicoin_dot_fun.move
```diff
@@ -2437,8 +2439,52 @@ module emoji::main {
...
+        market_address: address,
+        emoji_bytes: vector<u8>,
+        extend_ref: ExtendRef,
+    }
+
+    #[resource_group = ObjectGroup]
+    struct Registry has key {
+        registry_address: address,
+        supported_emojis: SmartTable<vector<u8>, u8>,
+        markets_by_emoji_bytes: SmartTable<vector<u8>, address>,
```
xbtmatt, line 2454: These are supposed to be the same keys right? Any reason not to use the market address as the value in `supported_emojis` instead?

xbtmatt: Where in the case it's not set yet it'd be `0x0` or something.? Or an `Option<address>` even

But I suppose it doesn't matter much, just duplicate keys

xbtmatt (approved):
Looks good to me! Just the one comment I made, but doesn't make much of a difference

src/move/sources/emojicoin_dot_fun.move
alnoki: You can create a market from an emoji combo, so this would only apply if we enforced a single emoji per market

xbtmatt: Oh yes of course...

PR #21, opened 2024-04-22
[ECO-1585] Add volume/TVL tracking, housekeeping

alnoki:
# Description

Add at market level:
1. Cumulative base volume
1. Cumulative quote volume

Add at global level, via aggregators (for parallelism):
1. Total quote locked
1. Cumulative quote volume

Housekeeping:
1. Fix an inconsistent instance of type args
1. Remove redundant type annotations
1. Replace `input_is_base` to more intuitive `is_sell`

# Testing

```sh
git ls-files | entr -c aptos move test --package-dir src/move/emojicoin_dot_fun --dev
```

# Checklist

- [x] ~Did you check off all checkboxes from the linked Linear task?

### DIFF
--- a/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
+++ b/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
@@ -462,7 +462,7 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
 
         // Burn coins by first withdrawing them from provider's coin store, to trigger event.
         let lp_coins = coin::withdraw<EmojicoinLP>(provider, event.lp_coin_amount);
-        burn_lp_coin<Emojicoin, EmojicoinLP>(market_ref_mut.market_address, lp_coins);
+        burn_lp_coins<Emojicoin, EmojicoinLP>(market_ref_mut.market_address, lp_coins);
 
         // Update market state.
         let reserves_ref_mut = &mut market_ref_mut.cpamm_real_reserves;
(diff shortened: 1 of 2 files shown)


src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -92,6 +93,8 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
         clamm_virtual_reserves: Reserves,
         cpamm_real_reserves: Reserves,
         lp_coin_supply: u128,
+        cumulative_base_volume: u128,
+        cumulative_quote_volume: u128,
```
xbtmatt, line 97: Just curious for my understanding since I've not used them much, why the usage of Aggregators in `Registry` but not per `Market`? I understand `Registry` will have significantly more resource contention but is there a downside to using aggregators, or is this mostly just an arbitrary decision?

xbtmatt (approved):
Overall looks good to me!

But just curious for my understanding since I've not used them much...why the usage of Aggregators in `Registry` but not for the cumulative quote/base in `Market`?

I understand Registry will have significantly more resource contention but is there a downside to using aggregators, or is this mostly just an arbitrary decision?

alnoki:
> Overall looks good to me!
> 
> But just curious for my understanding since I've not used them much...why the usage of Aggregators in `Registry` but not for the cumulative quote/base in `Market`?
> 
> I understand Registry will have significantly more resource contention but is there a downside to using aggregators, or is this mostly just an arbitrary decision?

Reading aggregators is more expensive than reading integers, and serializes txns anyways

At the market level txns will have to be serial since they modify the liquidity amounts so using an aggregator wouldn't help because we'd need to read during runtime

But at the global level they can be parallel, because we don't need to read during runtime. Instead we can use an aggregator snapshot in the events, for parallelism

And

PR #22, opened 2024-04-23
[ECO-1587] Add the rest of coin factory e2e testing

xbtmatt:
# Description

- [x] Make package mutable
- [x] Verify market registration
    - [x] Package w/ type args published, etc
    - [x] Verify structs are published as needed
- [x] Replace @coin_factory with @market_address
- [x] Then try registering a market
- [x] Then try swapping
    - [x] Just to verify that the types and coins are initialized, since swapping follows up and registers them
- [x] Make it have the LP-<market ID> for the symbol on LP token
    - [x] Update architecture spec with this
- [x] Consolidate contract publication to before/after tests so you can re-use the contract address instead of having to re-publish for all different unit/e2e tests
- [x] Add a #[test_only] module for a specific emoji (Yellow heart) and use it in tests with the properly derived market address
    - [x] Write a unit test to assert!(derived_yellow_heart_address == @yellow_heart_market_address, 0)
- [x] Organize layout under src/typescript
    - [x] Consolidate CI code to be under src/typescript/tests
    - [x] Then can have src/typescript/frontend

# Testing

All TypeScript tests are tested in CI, you can also test them like so:

```shell
cd src/typescript/api
pnpm i && pnpm run test
```

# Checklist

- [x] ~Did you check off all checkboxes from the linked Linear task?~


src/typescript/jest.config.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
     "src/utils/env.ts",
   ],
   testPathIgnorePatterns: ["dist/*"],
-  collectCoverage: true,
+  collectCoverage: false,
```
xbtmatt, line 15: Until we have more tests the jest coverage will just repeatedly fail. Just disabling for now, will try to re-enable later.

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -318,6 +326,7 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
             market_address
         );
         let swapper_address = signer::address_of(swapper);
+        assert!(account::exists_at(swapper_address), E_SWAPPER_DOES_NOT_EXIST);
```
xbtmatt, line 329: @alnoki so I added this here because on line `330` previously, the function call fails if the account already exists, which it will the vast majority of the time. I added an error code `E_SWAPPER_DOES_NOT_EXIST` and a simple test for it.

The error is thrown only in `swap` and just prior to simulating the swap. We could also put it inside the swap simulation function but that's your call.

Perhaps it's overkill, but I think it's the cleanest way to handle it.

xbtmatt:
@alnoki I can't remember if we had already addressed `Set clamm reserve values at beginning as needed per [ECO-1557](https://linear.app/econia-labs/issue/ECO-1557/implement-coin-type-module-factory) unaddressed comment` or not. If we haven't let me know otherwise feel free to just resolve this comment. I crossed it off in the checklist but didn't mark it as done yet.

doc/architecture-specification.md
```diff
@@ -86,7 +86,13 @@ These keywords `SHALL` be in `monospace` for ease of identification.
...
 1. Each Emojicoin's `symbol` field `SHALL` use hex-encoded UTF8.
 1. Each Emojicoin's `name` field `SHALL` use the format `<symbol> emojicoin`
-   where `<symbol>` is identical to the bytes from the `symbol` field.
+   where `<symbol>` is identical to the bytes from its `symbol` field.
+1. Each Emojicoin `SHALL` have a unique liquidity provider coin (LP coin)
+   associated with it.
+1. Each LP coin's `symbol` field `SHALL` use the format `LP-<market ID>`
+   where `<market ID>` is the corresponding market's `market_id`.
+1. Each LP coin's `name` field `SHALL` use the format `<symbol> emojicoin`
+   where `<symbol>` is identical to the bytes from its `symbol` field.
```
xbtmatt, line 95: Is this what we want to do btw @alnoki ?

Aka `LP-1` and `LP-1 emojicoin`

alnoki: E.g. `LP-3` and `<Smiley emoji bytes> emojicoin LP`

src/move/emojicoin_dot_fun/Move.toml
```diff
@@ -1,5 +1,6 @@
 [addresses]
 emojicoin_dot_fun = "_"
+yellow_heart_market_address = "0x6ba9e8b95825fffbfcc3fc743ec7461c13f334f8b3d5c0a30988b6d4399a003a"
```
alnoki, line 3: ?

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
alnoki:
```suggestion
        account::create_account_if_does_not_exist(swapper_address);
        coin::register<AptosCoin>(swapper);
```

alnoki: Assuming you're fine with this suggestion:

- [ ] Remove the error code since it is no longer needed

src/move/emojicoin_dot_fun/Move.toml
alnoki: Ahh okay I see what is going on here

How about moving this to `dev-addresses` with a comment explaning how this is the resulting address when getting a named object from the `emojicoin_dot_fun` dev-address?

alnoki: Actually, on second though, I think we should remove this and determine it programatically at runtime

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -816,6 +851,51 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
+        init_module(deployer);
+        let symbol_bytes = YELLOW_HEART;
+        let registry_ref_mut = borrow_registry_ref_mut();
+        let (market_address, _) = create_market(registry_ref_mut, symbol_bytes);
+        assert!(@yellow_heart_market_address == market_address, 0);
+
+        let user_addr = signer::address_of(user);
+        let input_amount: u64 = 100_000_000;
+        fund_account(user_addr, input_amount, aptos_framework);
+        let integrator_address = @0xbeefcafe;
```
alnoki, line 870:
```suggestion
        let integrator_address = @0xf00dcafe;
```

alnoki: More inclusive to cows

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -816,6 +851,51 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
+        swap<TestEmojicoin, TestEmojicoinLP>(
+            @yellow_heart_market_address,
+            user,
+            input_amount,
+            false, // Buy the emojicoins.
+            integrator_address,
+            0,
+        );
+
+        assert!(exists<LPCoinCapabilities<TestEmojicoin, TestEmojicoinLP>>(@yellow_heart_market_address), 0);
```
alnoki, line 880:
```suggestion
        assert!(
            exists<LPCoinCapabilities<TestEmojicoin, TestEmojicoinLP>>(
                @yellow_heart_market_address
            ),
        0);
```

alnoki: For 100 line break

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -816,6 +851,51 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
+    }
+
+    #[test(deployer=@emojicoin_dot_fun, user=@0xAA), expected_failure(abort_code=E_SWAPPER_DOES_NOT_EXIST)]
```
alnoki, line 883:
```suggestion
    #[test(deployer=@emojicoin_dot_fun, user = @0xAA)]
    #[expected_failure(abort_code = E_SWAPPER_DOES_NOT_EXIST)]
```

alnoki: Line breaking, ` = ` consistency with other tests

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -816,6 +851,51 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
+        deployer: &signer,
+        user: &signer,
+    ) acquires Registry, RegistryAddress, Market, LPCoinCapabilities {
+        initialize_module_and_yellow_heart_coins(deployer);
+        swap<TestEmojicoin, TestEmojicoinLP>(
+            @yellow_heart_market_address,
+            user,
+            1,
+            false,
+            @0xbeefcafe,
```
alnoki, line 894:
```suggestion
            @0xf00dcafe,
```

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -940,17 +1012,42 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
+    #[test(deployer=@emojicoin_dot_fun)]
+    fun test_coin_names_and_symbols(
+        deployer: &signer,
+    ) acquires Market, Registry, RegistryAddress {
+        initialize_module_and_yellow_heart_coins(deployer);
+
+        let symbol = string::utf8(YELLOW_HEART);
+        assert!(coin::symbol<TestEmojicoin>() == symbol, 0);
+        assert!(coin::name<TestEmojicoin>() == get_concatenation(symbol, string::utf8(EMOJICOIN_NAME_SUFFIX)), 0);
+        assert!(coin::name<TestEmojicoin>() == get_concatenation(symbol, string::utf8(b" emojicoin")), 0);
```
alnoki, line 1046: Line break at 100

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -940,17 +1012,42 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
+        let lp_symbol = string::utf8(b"LP-1");
+        let lp_name = string::utf8(b"LP-1 emojicoin");
+        assert!(coin::symbol<TestEmojicoinLP>() == lp_symbol, 0);
+        assert!(coin::name<TestEmojicoinLP>() == get_concatenation(lp_symbol, string::utf8(EMOJICOIN_LP_NAME_SUFFIX)), 0);
```
alnoki, line 1050: Line break at 100

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -940,17 +1012,42 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
-    struct EmojicoinLP {}
+    #[test(deployer=@emojicoin_dot_fun)]
+    fun test_coin_names_and_symbols(
+        deployer: &signer,
+    ) acquires Market, Registry, RegistryAddress {
+        initialize_module_and_yellow_heart_coins(deployer);
+
+        let symbol = string::utf8(YELLOW_HEART);
+        assert!(coin::symbol<TestEmojicoin>() == symbol, 0);
+        assert!(coin::name<TestEmojicoin>() == get_concatenation(symbol, string::utf8(EMOJICOIN_NAME_SUFFIX)), 0);
```
alnoki, line 1045: Line break at 100

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -270,11 +276,13 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
             coin::destroy_mint_cap(mint_cap);
             coin::destroy_burn_cap(burn_cap);
 
+            let market_id_str = string_utils::to_string(&market_ref.market_id);
+            let lp_symbol = get_concatenation(string::utf8(EMOJICOIN_LP_SYMBOL_PREFIX), market_id_str);
```
alnoki, line 280: Line break at 100

src/move/emojicoin_dot_fun/Move.toml
alnoki: We could event pass this in using `--named-addresses`?

There's just something about hardcoding this on here, as derived from another `dev-address`, that smells funny

alnoki (commented):
Major takeaways

1. I think we might want to figure out a different way to do the yellow heart address than hardcoding in Move.toml: I left some suggestions in code
1. I left a suggestion on the spec about naming LP coins that might require some Move refactors
1. I'm not in the best position to review the TS code, so I don't have any suggestions there

alnoki (changes requested):
Getting some issues with GitHub here and it's saying I need to leave a comment in this box to proceed with the review

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
xbtmatt: Okay I figured we'd probably want to avoid doing the error code, this would be great but I think `create_account_if_does_not_exist` is private unfortunately :( we can just add our own inline for it?

alnoki:
Per eng call, regarding the yellow heart address issue:

The behavior that needs to be tested is the generation of a market and coin types at a deterministic address, based on the `@emojicoin_dot_fun` address, e.g. named objects

This behavior could be tested purely in e2e, which would alleviate the limitations/confusion of hardcoding a derived address in a `Move.toml` file, which has a dependency on another `dev-address`

Note that a barrier is in `ensure_coins_initialized`, which gets the market address from a `&Ref Market`

To fix this:

- [ ] Create a test-only flow that allows you to mutate the market's address field, so that when the coin types are looked up, they can be verified under a test-only address like `0xaaa`

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
alnoki: We resolved this in #24

doc/architecture-specification.md
```diff
@@ -86,7 +86,13 @@ These keywords `SHALL` be in `monospace` for ease of identification.
    constant product calculations.
 1. Each Emojicoin's `symbol` field `SHALL` use hex-encoded UTF8.
 1. Each Emojicoin's `name` field `SHALL` use the format `<symbol> emojicoin`
-   where `<symbol>` is identical to the bytes from the `symbol` field.
+   where `<symbol>` is identical to the bytes from its `symbol` field.
+1. Each Emojicoin `SHALL` have a unique liquidity provider coin (LP coin)
```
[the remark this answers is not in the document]
xbtmatt: Done, posting here to make sure I did it correctly, added it to line 96 as well:

```markdown
1. Each `emojicoin`'s `symbol` field `SHALL` use hex-encoded UTF8.
1. Each `emojicoin`'s `name` field `SHALL` use the format `<symbol> emojicoin`
   where `<symbol>` is identical to the bytes from its `symbol` field.
1. Each `emojicoin` `SHALL` have a unique liquidity provider coin (LP coin)
   associated with it.
1. Each LP coin's `symbol` field `SHALL` use the format `LP-<market ID>`
   where `<market ID>` is the corresponding market's `market_id`.
1. Each LP coin's `name` field `SHALL` use the format `<symbol> emojicoin LP`
   where `<symbol>` is identical to the bytes from the corresponding
   `emojicoin`'s `symbol` field.
```

xbtmatt:
> Per eng call, regarding the yellow heart address issue:

So unfortunately the Move compiler doesn't check that module names are at different packages, so it thinks there could be a name collision for any modules that are named the same thing as long as it's all being compiled within the same package. This doesn't really make sense to me tbh because I got around it by putting them in a different package (meaning the logic is actually a bit flawed...?)

Here's the error message:

```json
"Error": "Unexpected error: Failed to run tests: Module and/or script names found that would cause failures on case insensitive file systems when compiling package 'EmojicoinDotFun':
The following modules and/or scripts would collide as 'coin_factory' on the file system:
    Module 'coin_factory' at path '/home/dev/emojicoin/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move'
    Module 'coin_factory' at path '/home/dev/emojicoin/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move'
    Module 'coin_factory' at path '/home/dev/emojicoin/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move'
Please rename these scripts and/or modules to resolve these conflicts."`
```

You can reproduce the error with this at the top of a file:
```rust
#[test_only] module 0xaaa::coin_factory {}
#[test_only] module 0xbbb::coin_factory {}
#[test_only] module 0xccc::coin_factory {}
```

It works if you only use one, but not more than one...so unfortunately I had to just add three coin factory packages with different module names. I managed to do it not using named addresses though, let me know what you think

alnoki (approved):

PR #25, opened 2024-04-25
[ECO-1595] add allowlist server, pre-commit hook

CRBl69:
# Description

This PR adds an allowlister web server to enable a private beta for emojicoins.

# Testing

To test this PR:

- create an `allowlist.txt` file
- add some addresses to that file
- set `ALLOWLIST_FILE` environment variable to the path of `allowlist.txt`
- run `cargo run -p allowlister3000`
- in another terminal, run `curl localhost:3000/0x…` (test some addresses that are in the file, some that are not)

# Checklist

- [x] ~~Did you update the changelog?~~
- [x] Did you check off all checkboxes from the linked Linear task? (Ignore if


.gitignore
```diff
@@ -1,3 +1,4 @@
 .env
 .env.local
 .env.development
+/target
```
alnoki, line 4: How about having this be specific to each related rust project?

Cargo.lock
alnoki, line 1: How about having this be specific to the particular project?

src/rust/allowlister3000/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Addresses must start with `0x`, and can have any amount of leading zeros.
+AllowLister3K does not check for invalid hex strings. AllowLister3K checks that
+incoming addresses over the network aren't bigger than 66 characters (`0x` +
+64). This check is not performed on the allowlist file.
+
+## HTTP requests
+
+To make a request to AllowLister3K, simply make a GET request to `/address`
+where `address` is the address in the correct format you want to check.
```
alnoki, line 30: How about some instructions here about launching this on GCP?

src/rust/allowlister3000/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+use std::{collections::HashSet, sync::Arc};
+use tower_http::cors::{Any, CorsLayer};
+
+fn sanitize(mut address: String) -> String {
+    loop {
+        if address == "0x" {
+            break String::from("0x0");
+        }
+        if address.starts_with("0x0") {
+            address.remove(2);
```
alnoki, line 16:
```suggestion
            address.remove(FIRST_HEX_CHAR);
```

src/rust/allowlister3000/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+            break address;
+        }
+    }
+}
+
+async fn allowlist(
+    Path(address): Path<String>,
+    State(state): State<Arc<HashSet<String>>>,
+) -> String {
+    if address.len() > 66 || !address.starts_with("0x") {
```
alnoki, line 27:
```suggestion
    if address.len() > MAX_ADDRESS_LENGTH || !address.starts_with("0x") {
```

src/sh/cargo-format.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+cargo fmt
```
alnoki, line 3: Will this recursively run if all Rust source is moved to `src/rust/allowlist` ?

.gitignore
CRBl69: That would not work. The target directory is located at `/target`. If I do what you're describing, it will only match `/src/rust/**/target`, which will not match `/target`.

Cargo.toml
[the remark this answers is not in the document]
CRBl69: I was using a Cargo workspace for ease of use. Usually, the `Cargo.toml` file to define the workspace is in root. See:

- https://github.com/aptos-labs/aptos-core
- https://github.com/denoland/deno
- https://github.com/alacritty/alacritty

Aptos is a good example: despite not being a 100% Rust based codebase, it still defines the Rust workspace at root (by having `/Cargo.toml`), just like it does for its NPM workspace (by having `/package.json`).

Generally, this is very helpful, for example, it allowed me to add the `pre-commit` hook that runs `cargo fmt` without having to `cd` into a rust specific directory. This means you can check everything from root (e.g. I could run `cargo check && cargo fmt`, `npm lint`, and any other thing from root).

src/rust/allowlister3000/README.md
CRBl69: Added.

src/rust/allowlister3000/src/main.rs
CRBl69: Done.

src/rust/allowlister3000/src/main.rs
CRBl69: Done.

CRBl69:
Note: could not make `pre-commit` checks pass since `mdformat` and `markdownlint-fix` conflict. I would suggest only using `mdformat`. Two is overkill, and there is no need for such surgical precision in our Markdown formatting IMO.

lol
alnoki, line 1: ?

alnoki (commented):
Per eng call

- [x] Move `Cargo.toml` in root to `src/rust`
- [ ] Update `cargo-format.sh` to use this
- [x] Use https://github.com/doublify/pre-commit-rust, with the 3 hooks listed on https://pre-commit.com/hooks.html (`clippy`, `fmt`, `cargo check`)

.gitignore
alnoki: Per https://github.com/econia-labs/emojicoin-dot-fun/pull/25#pullrequestreview-2025533633, take out top-level gitignore since won't be needed

alnoki (approved):

PR #35, opened 2024-05-01
[ECO-1581] Add emoji chat feature

xbtmatt:
# Description

 - Pass in a string of emojis
 - They all get validity checked
 - Expand support with chat emojis
 - Filter out the largest/most egregious emojis with excessive byte sizes
 - Optimize the emoji chat message constructor by caching or using index sequencing (we used the latter)
 - Limit message size to 100 emojis
 - Waive market registration fee is user is initializing chat emojis
 - Update TypeScript unit tests accordingly

# Testing

Both the Move unit tests and TypeScript unit tests have been updated

# Checklist


src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -411,7 +447,17 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
             assign_supply_minuend_reserves_ref(market_ref_mut, in_bonding_curve);
         let circulating_supply = supply_minuend - reserves_ref.base;
-        let user_emojicoin_balance = coin::balance<Emojicoin>(user_address);
+        let user_emojicoin_balance = if (!coin::is_account_registered<Emojicoin>(user_address)) {
+            coin::register<Emojicoin>(user);
+            0
+        } else {
+            coin::balance<Emojicoin>(user_address)
+        };
+        let balance_as_fraction = if (circulating_supply == 0) {
```
xbtmatt, line 456: Highlighting this to flag it @alnoki , this was giving me divide by 0 error. It's possible I just didn't initialize things correctly in my unit tests, but not sure.

alnoki: It's possible that denominator is 0 if no coins are in circulation yet

e.g. you register market then immediately chat

should probably just escape that case, I'll catch it in tests

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -1169,6 +1214,36 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
+        } else {
+            num_chat_emojis
+        };
+
+        // The next index to add is equal to the length of the smart table.
+        for (i in next_idx..last_idx) {
+            let emoji_bytes = *vector::borrow(&chat_emojis, i);
+            smart_table::add(chat_emojis_ref_mut, emoji_bytes, 0);
+        };
+    }
```
alnoki, line 1245: I think it's not a problem for the first market registrant to pay I/O overhead of stuffing this into a table, rather than smart table

Makes things cleaner

Then we just need to stich into register market function, and all emojis will be initted by the time the first market is ready to go

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -386,6 +401,27 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
+
+        // Construct the chat message from the emoji indices.
+        let message: String = string::utf8(b"");
+        vector::for_each(emoji_indices_sequence, |idx| {
+            let emoji = *vector::borrow(&emoji_bytes, (idx as u64));
+            string::append_utf8(&mut message, emoji);
+        });
+
+        assert!(
+            string::length(&message) <= (MAX_CHAT_MESSAGE_LENGTH as u64),
```
alnoki, line 421: This seems pretty restrictive, I estimate that the cap will be only 15/20 emojis in some cases?

How about capping the length of `emoji_indices_sequence` to `MAX_CHAT_MESSAGE_LENGTH`?

src/move/emojicoin_dot_fun/sources/hex_codes.move
```diff
@@ -2313,6 +2313,1481 @@ module emojicoin_dot_fun::hex_codes {
...
+            x"f09fa791e2808df09f8db3",                                                  // Cook [1F9D1 200D 1F373]
+            x"f09fa791f09f8fbfe2808df09f8db3",                                          // Cook: dark skin tone [1F9D1 1F3FF 200D 1F373]
+            x"f09fa791f09f8fbbe2808df09f8db3",                                          // Cook: light skin tone [1F9D1 1F3FB 200D 1F373]
+            x"f09fa791f09f8fbde2808df09f8db3",                                          // Cook: medium skin tone [1F9D1 1F3FD 200D 1F373]
+            x"f09fa791f09f8fbee2808df09f8db3",                                          // Cook: medium-dark skin tone [1F9D1 1F3FE 200D 1F373]
+            x"f09fa791f09f8fbce2808df09f8db3",                                          // Cook: medium-light skin tone [1F9D1 1F3FC 200D 1F373]
+            x"f09f91a8e2808de29da4efb88fe2808df09f91a8",                                // Couple with heart: man, man [1F468 200D 2764 FE0F 200D 1F468]
+            // x"f09f91a8f09f8fbfe2808de29da4efb88fe2808df09f91a8f09f8fbf",                // Couple with heart: man, man, dark skin tone [1F468 1F3FF 200D 2764 FE0F 200D 1F468 1F3FF]
```
alnoki, line 2339: Are we able to uncomment these if we go with this approach?

https://github.com/econia-labs/emojicoin-dot-fun/pull/35#discussion_r1587728380

.gitignore
```diff
@@ -0,0 +1 @@
+.tmp/
```
alnoki, line 1: Remove?

src/python/move_emojis/utils/data_parser.py
alnoki, line 1: Include logic that reduces excessively chained sequences

- [ ] COMMENT WHICH ONES ARE ALLOWED AND IF THEY ARE ELSEWHERE
- [ ] Include the base cases that are allowed as constants, and eliminate all that chain on top of those
- [ ] Note that nothing which is a symbol has changed, for example you might need to escape men holding hands because it is already a symbol

src/move/emojicoin_dot_fun/sources/hex_codes.move
alnoki: Noting here that these are disabled due to package size restrictions, not execution limits

The Python script should be updated to generate this and eliminate permutations in the case of certain long bases

.gitignore
alnoki: If possible, put in a CI-specific directory

xbtmatt: Done!

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
xbtmatt: Done, only thing to note is that I just set the `fee` to 0 for now, although not sure we want to do that, but as noted in Slack there's no way to know the gas fees at runtime

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
xbtmatt: Done!

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -461,18 +504,24 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
...
-        aptos_account::transfer(registrant, integrator, MARKET_REGISTRATION_FEE);
+        let fee = if (ensure_chat_emojis_initialized(registry_ref_mut)) {
+            0
+        } else {
+            let can_pay_fee =
+                coin::is_account_registered<AptosCoin>(registrant_address) &&
+                coin::balance<AptosCoin>(registrant_address) >= MARKET_REGISTRATION_FEE;
+            assert!(can_pay_fee, E_UNABLE_TO_PAY_MARKET_REGISTRATION_FEE);
+            aptos_account::transfer(registrant, integrator, MARKET_REGISTRATION_FEE);
+            (MARKET_REGISTRATION_FEE as u128)
```
alnoki, line 519: Since now the fee is not necessarily static, how about an `integrator_fee_paid` field in the `MarketRegistrationEvent`

Can be 0 for the first registration, else `MARKET_REGISTRATION_FEE` for all others

Just useful for accounting

src/python/move_emojis/utils/data_parser.py
xbtmatt: Explained it in the function docstring, should be sufficient lmk

src/move/emojicoin_dot_fun/sources/hex_codes.move
xbtmatt: Done!

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
alnoki: Marking resolved, slated for unit testing

alnoki (approved):

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
alnoki: Addressed at 3077b86 on #37

PR #37, opened 2024-05-02
ECO-1571

alnoki:
# Description

Delete this line and add a description that explains what changes you have done
and why they were necessary.

# Testing

PR.

# Checklist

- [ ] Did you check off all checkboxes from the linked Linear task? (Ignore if

```md
```


src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -1260,9 +1263,10 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
             },
         };
 
-        // Load supported emojis into registry.
+        // Load supported coin symbol emojis into registry.
+        let coin_symbol_emojis_ref_mut = &mut registry.coin_symbol_emojis;
```
alnoki, line 1267: Would need to look more into the disassembled bytecode, but I have a hunch doing this declaration before the `for_each_ref` loop will avoid the gas cost associated with field access that was previously inside each iteration (e.g. assuming compiler isn't optimizing this out)

xbtmatt: Oh interesting- you're saying you think each loop iteration is charged for a field access if you use `&mut registry.coin_symbol_emojis` as opposed to `let coin_symbol_emojis_ref_mut = &mut registry.coin_symbol_emojis;` before the loop?

alnoki: That's my hunch

Before a few days ago I would've assumed that it is optimized out

But now I suspect the compiler has no optimizations

src/move/emojicoin_dot_fun/sources/tests.move
```diff
@@ -0,0 +1,546 @@
...
+        // Test unsupported emojis.
+        assert!(!is_a_supported_symbol_emoji(x"0000"), 0);
+        assert!(!is_a_supported_symbol_emoji(x"fe0f"), 0);
+        assert!(!is_a_supported_symbol_emoji(x"1234"), 0);
+        assert!(!is_a_supported_symbol_emoji(x"f0fabcdefabcdeff0f"), 0);
+        assert!(!is_a_supported_symbol_emoji(x"f0f00dcafef0"), 0);
+        // Minimally qualified "head shaking horizontally".
+        assert!(!is_a_supported_symbol_emoji(x"f09f9982e2808de28694"), 0);
+
+        // Verify a supported emoji, add some bunk data to it, then verity it is no longer allowed.
```
xbtmatt, line 505:
```suggestion
        // Verify a supported emoji, add invalid data to it, then verify it is no longer allowed.
```

looks like I made a typo 😅

xbtmatt (approved):
Looked it over, it's much cleaner this way, and thanks for cleaning up some of my stuff...

I also ran the tests and the code generation myself, LGTM!

PR #38, opened 2024-05-02
ECO-1628 Fix the Launch Emojicoin page

maksymantonenko:
# Description

Added the ability to select multiple emojis
Added validation for a 10-byte limit
Added lazy loading for the emoji palette

# Testing

PR.

# Checklist

- [ ] Did you check off all checkboxes from the linked Linear task? (Ignore if

```md
```

### DIFF
--- a/src/typescript/frontend/src/pages/launch-emojicoin/index.tsx
+++ b/src/typescript/frontend/src/pages/launch-emojicoin/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect } from "react";
 import EmojiPicker, { Theme, EmojiClickData } from "emoji-picker-react";
 
 import { useThemeContext, useTranslation } from "context";
@@ -14,10 +14,10 @@ const LaunchEmojicoinPage: React.FC = () => {
   const { t } = useTranslation();
   const { theme } = useThemeContext();
 
-  const [emoji, setEmoji] = useState<EmojiClickData | null>(null);
+  const input = document.querySelector("input");
 
   const { validationSchema, initialValues } = useValidationSchema();
-  const { errors, touched, fieldProps, setFieldValue } = useForm({
+  const { values, errors, touched, fieldProps, setFieldValue } = useForm({
     initialValues,
     validationSchema,
     enableReinitialize: true,
@@ -26,12 +26,20 @@ const LaunchEmojicoinPage: React.FC = () => {
     },
   });
 
+  const names = values.emojiList.map(emoji => emoji.names[0]).join(",");
+  const tickers = values.emojiList.map(emoji => emoji.emoji).join(",");
+
   const { targetRef: targetRefLabel, tooltip: labelTooltip } = useTooltip(
     t("Pick one to five emojis; due to byte limitations not all combinations are supported."),
     {
       placement: "top",
     },
   );
+
+  const { targetRef: targetRefEmojiName, tooltip: tooltipEmojiName } = useTooltip(undefined, {
+    placement: "top",
+    isEllipsis: true,
+  });
 
   const { targetRef, tooltip } = useTooltip(
     <StyledEmojiPickerWrapper>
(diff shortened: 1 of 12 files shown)


.idea/workspace.xml
xbtmatt, line 1: Going to remove this @maksymantonenko, assuming you didn't mean to add it

.idea/prettier.xml
xbtmatt, line 1: Same as this

src/typescript/frontend/src/pages/launch-emojicoin/index.tsx
```diff
@@ -14,10 +14,10 @@ const LaunchEmojicoinPage: React.FC = () => {
   const { t } = useTranslation();
   const { theme } = useThemeContext();
 
-  const [emoji, setEmoji] = useState<EmojiClickData | null>(null);
+  const input = document.querySelector("input");
```
xbtmatt, line 17:
```suggestion
  const input = inputRef.current;
```

and later:

```typescript
  // later, <Input ... ref={inputRef} />
```

Can we convert this to a ref with the input element, ideally we'd adhere to React best practices, but let me know if perhaps there's more going on here that I'm not understanding

src/typescript/frontend/src/pages/launch-emojicoin/index.tsx
```diff
@@ -50,11 +59,40 @@ const LaunchEmojicoinPage: React.FC = () => {
...
   const onEmojiClickHandler = async (emoji: EmojiClickData) => {
-    await setFieldValue("emoji", emoji.emoji);
-    setEmoji(emoji);
-    console.log("emoji", emoji);
+    await setFieldValue("emoji", values.emoji + emoji.emoji);
+    await setFieldValue("emojiList", [...values.emojiList, emoji]);
+  };
+
+  const inputProhibition = async (event: KeyboardEvent) => {
+    if (event.key !== "Backspace") {
```
xbtmatt, line 67: Ideally we'd allow for arrow key navigation here as well (left/right)

src/typescript/frontend/src/pages/launch-emojicoin/index.tsx
```diff
@@ -50,11 +59,40 @@ const LaunchEmojicoinPage: React.FC = () => {
...
+      const newEmojiListValue = values.emojiList.slice(0, -1);
+      await setFieldValue("emoji", newEmojiValue);
+      await setFieldValue("emojiList", newEmojiListValue);
+    }
   };
 
+  const cursorPositioning = () => {
+    input!.focus();
+    input!.selectionStart = input!.value.length;
+  };
```
xbtmatt, line 80:
```suggestion
  const cursorPositioning = () => {
    if (input) {
      input.focus();
      input.selectionStart = input.value.length;
    }
  };
```

src/typescript/frontend/src/pages/launch-emojicoin/index.tsx
```diff
@@ -50,11 +59,42 @@ const LaunchEmojicoinPage: React.FC = () => {
...
+      input.addEventListener("click", cursorPositioning);
+    }
+
+    return () => {
+      if (input) {
+        input.removeEventListener("keydown", inputProhibition);
+        input.removeEventListener("click", cursorPositioning);
+      }
+    };
+  }, [input, values]);
```
xbtmatt, line 96: I think you can actually move all the `input` stuff, aka `inputProhibition`, `cursorPositioning`, `eventListeners` etc inside of the `useEffect`

If you refactor to an `inputRef` as well you can put `const input = inputRef.current` in there, too

xbtmatt (changes requested):
1. Backspace should remove the entire emoji, not just parts of it (you'll probably need some special parsing function when backspace is hit to remove the full emoji).
2. This may also apply to the arrow keys as well, which we should allow as input
3. Remove the JetBrains .idea workspace files
4. Refactor `document.querySelector` if possible

xbtmatt (approved):
Thanks for the quick turnaround, looks good!

PR #41, opened 2024-05-04
[ECO-1629] Test module/market init to 100% coverage

alnoki:
# Description

## Production code

1. Since the [coverage tool is broken for inline functions][coverage tool], take
   out all inline annotations except where:
    1. The function returns a borrowed reference.
    1. The function would require an `acquires` statement if not inline.
1. Add swap/buy trigger differentiation on swaps.
1. Track market-wise Aptos Coin balance in a `MarketView`.
1. Update initial market stats/state triggers.
1. Update tracking for market registration fees.
1. Add inline FDV helper function.
1. Fix start time/period return mismatch in periodic state tracker decoder.

## Test code

1. Add assorted decoders for test-only state to main module.
1. Add struct mocking with base values, and struct-wise assertion.
1. Add tests to assert all state (global storage, emitted events) modifications
   during package publication and market registration.
1. Ensure 100%

## Miscellaneous

1. Clarify nonce vs bump time in architecture specification.

# Testing

From in `src/move/emojicoin_dot_fun`

```sh
git ls-files | entr -c aptos move test --coverage --dev
```

Compare coverage against bytecode:

```sh
aptos move coverage bytecode --module emojicoin_dot_fun --dev
```

[coverage tool]: https://github.com/aptos-labs/aptos-core/issues/9154

# Checklist

### DIFF
--- a/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
+++ b/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
@@ -2257,18 +2257,6 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
 
     #[test_only] public fun init_module_test_only(account: &signer) { init_module(account) }
 
-    #[test_only] public fun pack_market_metadata(
-        market_id: u64,
-        market_address: address,
-        emoji_bytes: vector<u8>,
-    ): MarketMetadata {
-        MarketMetadata {
-            market_id,
-            market_address,
-            emoji_bytes,
-        }
-    }
-
     #[test_only] public fun pack_reserves(base: u64, quote: u64): Reserves {
         Reserves { base, quote }
     }
--- a/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
+++ b/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
@@ -2251,6 +2251,9 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
     #[test_only] public fun get_PERIOD_1H(): u64 { PERIOD_1H }
     #[test_only] public fun get_PERIOD_4H(): u64 { PERIOD_4H }
     #[test_only] public fun get_PERIOD_1D(): u64 { PERIOD_1D }
+    #[test_only] public fun get_REGISTRY_NAME(): vector<u8> { REGISTRY_NAME }
+    #[test_only] public fun get_TRIGGER_MARKET_REGISTRATION(): u8 { TRIGGER_MARKET_REGISTRATION }
+    #[test_only] public fun get_TRIGGER_PACKAGE_PUBLICATION(): u8 { TRIGGER_PACKAGE_PUBLICATION }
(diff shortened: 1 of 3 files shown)


src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
```diff
@@ -601,7 +618,7 @@ module emojicoin_dot_fun::emojicoin_dot_fun {
                 base_volume: 0,
                 quote_volume: 0,
                 nonce: 0,
-                time,
+                time: 0,
```
xbtmatt, line 621: so this is `0` because `last_bump_time` records the current time now, right? What exactly does `time` track now?

src/move/emojicoin_dot_fun/sources/tests.move
```diff
@@ -94,6 +142,170 @@
...
+        instantaneous_stats: TestInstantaneousStats,
+        last_swap: TestLastSwap,
+    }
+
+    struct TestStateMetadata has copy, drop, store {
+        market_nonce: u64,
+        bump_time: u64,
+        trigger: u8,
+    }
+
```
xbtmatt, line 308: What's the reasoning behind duplicating all these Test structs? Is there a dependency issue or something? Just wondering because duplication of this logic is probably error prone, but I assume the compiler is just being a pain again..? 😓

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
alnoki: This field tracks the time of the last swap

It initializes to 0, basically as a null value 

(it would be more expensive to use an option here, though that might be slightly more intuitive)

src/move/emojicoin_dot_fun/sources/tests.move
alnoki: I've chosen to use struct here for mocking so that we can enforce named fields

When testing Econia v4, it ended up that I had to write a bunch of assert statements for each individual value that was changed

As a result there ended up being a ton of copy-paste, which is not so maintainable when one thing changes and you have to go through and update every test

Hence to abstract everything there is now a suite of `base_<struct_name>` functions, which return default values

Then during a test you only need to manually modify certain fields that deviate from the base, and call `assert_<struct_name>` which does all the unpacking internally

Check out `register_market_comprehensive_state_assertion()`, which would be way more lines if we were manually asserting everything and not relying on named mock fields

src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move
xbtmatt: Ahh okay, no that makes sense, just was confused why it went from `time` to `0`

xbtmatt (approved):

src/move/emojicoin_dot_fun/sources/tests.move
alnoki: Yes this is correct

I've updated other followup testing task ECO-1640 to update to `Mock` and also add some docs

Will merge this so APIs are available in `main`, then address those two items in next testing PR

PR #43, opened 2024-05-07
[ECO-1642] Pools page

maksymantonenko:
# Description

Refactored elements in the header.
Pools Page


xbtmatt (approved):
Awesome, looks great. :) I'll leave you some more feedback on slack

PR #44, opened 2024-05-07
[ECO-1645] Add multisig publication instructions

alnoki:
# Description

Add to the package README instructions on publishing onchain using a multisig.

# Testing

Published onchain using a multisig.

# Checklist

### DIFF
--- a/src/move/emojicoin_dot_fun/README.md
+++ b/src/move/emojicoin_dot_fun/README.md
@@ -1,7 +1,9 @@
 # emojicoin dot fun
 
 This package is just within the bounds of max publication gas, and cost roughly
-1 APT to publish during testing. To publish:
+1 APT to publish during testing.
+
+## Signer-signer publication
 
 ```sh
 aptos move publish \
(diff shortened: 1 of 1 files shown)


src/move/emojicoin_dot_fun/README.md
```diff
@@ -1,7 +1,9 @@
 # emojicoin dot fun
 
 This package is just within the bounds of max publication gas, and cost roughly
-1 APT to publish during testing. To publish:
+1 APT to publish during testing.
```
xbtmatt, line 4:
```suggestion
1 APT to publish during testing.

Keep in mind that the account sending the publish transaction needs to hold
more APT if you set the max gas to more than 1 APT.

For example, at `---max-gas 2000000` with `--gas-unit-price 100`, your account
needs to have at least 2 APT to be able to publish, even if it only costs 1 APT.
```

xbtmatt: (or something like this)

xbtmatt (commented):
Sweet, works for me, just a few changes and it's solid

xbtmatt (approved):

PR #48, opened 2024-05-11
[ECO-1659]

alnoki:
# Description

Delete this line and add a description that explains what changes you have done
and why they were necessary.

# Testing

PR.

# Checklist

- [ ] Did you check off all checkboxes from the linked Linear task?
  (Ignore if you are not a member of Econia Labs)

```md
```


src/move/emojicoin_dot_fun/sources/tests.move
```diff
@@ -1469,6 +1508,400 @@
...
+                    clamm_virtual_reserves = MockReserves {
+                        base: clamm_virtual_reserves.base - base_volume,
+                        quote: clamm_virtual_reserves.quote + quote_volume,
+                    };
+                }
+            };
+            net_proceeds = base_volume;
+            market_aptos_coin_balance = setup_market_view.aptos_coin_balance + quote_volume;
+            market_emojicoin_balance = setup_market_view.emojicoin_balance - base_volume;
+        };
```
xbtmatt, line 1689: Is this all duplicated logic from `simulate_swap_inner` just to recreate the mock data? I think all of the duplicated logic here just for the test setup is a sign that the function logic should be broken up

I mean I understand that the unit test framework isn't great for test setups and stuff, but it's difficult to tell here what exactly is being tested vs what's being duplicated

alnoki: > Is this all duplicated logic from `simulate_swap_inner`

Not copy-pasted, but yes the same logic (I reimplemented without looking at
source code, then verified that both had same output)

The catch is that without something like this, I'd be writing 9 or so individual
tests with a bunch of the same code blocks. Then if something in the core logic
changed, we'd have to update the manual code for each test, so not very
maintainable.

I'm trying to get enough abstraction to ease testing, but still to have the
testing be simple enough to ensure that the verification is happening in a
straightforward manner. Hence I agree with:

> sign that the function logic should be broken up

In the ideal case, I think we would:

1. Write core source code in a maximally-abstracted manner, then test all the
   individual helper functions on their own.
1. Write tests to use a central harness flow like the one I've added here, but
   rely mainly on the already-verified helper functions.
1. Test everything to 100% coverage with full state mutation assertions on each
   test.
1. Manually optimize/inline the core code, while ensuring tests still pass.

Unfortunately, it looks like we went straight to the final step rather than
optimizing at the end, so there aren't as many of these decomposed helpers
available for testing.

I'll start implementing some and see how we can make the test harness more
elegant.

alnoki: Per eng call updated to

https://github.com/econia-labs/emojicoin-dot-fun/blob/75909f311ce684031191738f648fe06c4af01b1a/src/move/emojicoin_dot_fun/sources/tests.move#L1611-L1640

xbtmatt (approved):
Awesome, appreciate you splitting these up, it's much more straightforward. Looks good to me!

src/move/emojicoin_dot_fun/sources/tests.move
xbtmatt: Thanks for doing this for lil old me, it's significantly easier for me to read and understand 😄

PR #71, opened 2024-05-20
[ECO-1697] Add the `charting_library` git submodule

xbtmatt:
# Description

- [x] Create the `static` folder in the repo and update the `.gitmodules` file
- [x] Update `inbox` `.gitmodules` to use `https` instead of `ssh`
- [x] Update this repo's `.gitmodules` in the same way
- [x] Update the `README.md` in `inbox`
- [x] Update the `README.md` in this repo
- [x] Add `tradingview` to the cspell dictionary

# Checklist

- [x] Did you update relevant documentation (if applicable)?
- [x] Did you add tests to cover new code or a fixed issue (if applicable)?
- [x] Did you update the changelog (if applicable)?
- [x] Did you check all checkboxes from the linked Linear task (if applicable)?
  (Ignore if you are not a member of Econia Labs)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "src/inbox"]
 	path = src/inbox
-	url = git@github.com:econia-labs/inbox.git
+	url = https://github.com/econia-labs/inbox.git
 [submodule "src/typescript/frontend/public/static"]
 	path = src/typescript/frontend/public/static
-	url = git@github.com:tradingview/charting_library.git
+	url = https://github.com/tradingview/charting_library.git
+	ignore = all
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,4 @@
 	url = git@github.com:econia-labs/inbox.git
 [submodule "src/typescript/frontend/public/static"]
 	path = src/typescript/frontend/public/static
-	url = https://github.com/tradingview/charting_library.git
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = git@github.com:econia-labs/inbox.git
 [submodule "src/typescript/frontend/public/static"]
 	path = src/typescript/frontend/public/static
-	url = https://github.com/tradingview/charting_library.git
+	url = git@github.com:tradingview/charting_library.git
```
alnoki, line 6: Is this relying on the private or public version?

If public, how about we keep it as `https` for accessibility?

(In multiple cases we've had downstream devs who aren't aware of SSH auth on GitHub and flag submodule issues)

xbtmatt: That's the private version, and yes I agree, I'll change them both to `https`, is there any reason we have `inbox` as ssh?

alnoki: So if it is the private version then unfortunately it'll have to use SSH

But we can use https for public repos

Please do go ahead and update it for inbox then

alnoki (approved):

src/inbox
alnoki, line 1: This should be updated in the inbox repo, not here

Then once the inbox PR lands we can bump the commit here

PR #133, opened 2024-06-28
[ECO-1902] Update pools page styling

CRBl69:
# Description

This PR updates the Pools page styling.

Noting that the responsive issue is fixed by ECO-1903.

# Testing

See Vercel preview.

# Checklist


xbtmatt (commented):
I'm getting two lines of text right between a screen width of 792px and 768px btw:

Screenshot 2024-06-28 at 9 01 58 PM

CRBl69:
@xbtmatt, that should not be a problem as `TOP 20/ALL` is removed in #131 and thus this does not happen.

xbtmatt (commented):
- Add the "you deposit" text to the pools input
- Make sure the emojis are centered/consistent for both the `pools` and `swap` page
- You can just use separate styling for each if necessary

alnoki (approved):

PR #136, opened 2024-06-28
[ECO-1812] mobile menu updates

CRBl69:
# Description

- [x] Add connect wallet button to mobile menu

# Testing

See vercel preview.

# Checklist


src/typescript/frontend/src/components/header/wallet-button/ConnectWalletButton.tsx
```diff
@@ -67,41 +69,82 @@ export const ButtonWithConnectWalletFallback: React.FC<ConnectWalletProps> = ({
...
+              </div>
+            </Button>
+            <Arrow width="18px" color="black" />
+          </StyledItemWrapper>
+        ) : (
+          children
+        )}
+      </>
+    );
+  }
```
xbtmatt, line 147: Instead of duplicating the component I think we can just do something like this:

https://github.com/econia-labs/emojicoin-dot-fun/blob/891f958cd4fa13d4cb5661eab22ede3466aac060/src/typescript/frontend/src/components/header/wallet-button/ConnectWalletButton.tsx#L62

Especially since after my merge a cople of things change here- I changed the voltage `p` elements to use `mt-2.5` for example so they're centered better

You mind if we just use the other code here? that is, the re-ordering so connect wallet is first on mobile?

xbtmatt: Here's the other part to that:

https://github.com/econia-labs/emojicoin-dot-fun/blob/4990c9f8bdbf04a354567b580397f55e62453653/src/typescript/frontend/src/components/header/wallet-button/ConnectWalletButton.tsx#L75

xbtmatt (approved):

PR #139, opened 2024-07-02
[ECO-1927] optimize queries and add loading pages

CRBl69:
# Description

Optimize queries by not using `wrap` which will make postgres use indexes.

Also adds loading pages.

### DIFF
--- a/src/typescript/frontend/src/app/home/loading.tsx
+++ b/src/typescript/frontend/src/app/home/loading.tsx
@@ -5,7 +5,5 @@ import React from "react";
 import LoadingComponent from "components/loading";
 
 export default function Loading() {
-    return (
-        <LoadingComponent />
-    );
+  return <LoadingComponent />;
 }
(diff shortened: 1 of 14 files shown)


xbtmatt (commented):
Awesome, this is significantly faster, it's a solid start. We discussed a few things so I'll just reiterate them here as next steps:

1. Store all of our static images on a CDN and deliver them from that instead of from `/static`
2. Change the `launch/page.tsx` to statically render the page and see how much it improves the load time. If it's a substantial difference, perhaps the dynamic render is the culprit, and we can look into that further

For other PRs:

3. Update the loading page to be custom for each component (this may require a separate PR, because it may require a decent time investment for each page)
4. We can consider storing the TradingView library on a CDN as well and delivering the js that way.


For now, we should ask @xorkiki if she's okay with the current Loading component or how we should make it generic so we can use it temporarily (or if we shouldn't)

xbtmatt (approved):

PR #156, opened 2024-07-09
[ECO-1939] change anoying liquidity popup to not be anoying anymore

CRBl69:
# Description

Add "i" button to avoid showing liquidity popup on page load and hide inputs.

# Testing

See vercel preview (pools page).

# Checklist


xbtmatt (commented):
Nice, much better!

For this though Screenshot 2024-07-10 at 2 20 00 AM

 I think this might be less intrusive and is pretty straightforward to code:
Screenshot 2024-07-10 at 2 21 01 AM

https://github.com/econia-labs/emojicoin-dot-fun/blob/4ae3e9b4580fc9bb5e6da41b561f1be54b626e77/src/typescript/frontend/src/components/emoji-picker/ColoredBytesIndicator.tsx#L26

CRBl69:
@xbtmatt I agree that the popup is annoying. I thought about your idea but wasn't sure where to place the indicators. I ended up putting your balance in each input. I did not use a `x/y` notation since you can already see your balance on the upper row and the amount on the lower row. Please let me know what you think about this.

xbtmatt (approved):
This looks great! Just two mall things:

1. Change the cursor on hover for the button if it's disabled (use `not-allowed` while balance is insufficient) that
2. You can conditionally style the balance color with green/red based on the balance being sufficient/insufficient

I think you should for sure do #1 but #2 could also help to make things less confusing but it's up to your discretion

Approving because it looks good to me!

PR #160, opened 2024-07-12
[ECO-1667] update routing system

CRBl69:
# Description

Updates the market routing system:

- market names are defined by the name of the emojis as follows: the name of the market 🍑🍑 is `peach;peach`.
- a market is accessible via `/market/MARKET_NAME`
- going to `/market/EMOJIS` will redirect to `/market/MARKET_NAME`.

For example, when accessing `https://emojicoin.fun/market/🍑🍑`, you will get redirected to `https://emojicoin.fun/market/peach,peach`.

# Testing

See vercel preview.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/app/market/[market]/page.tsx
+++ b/src/typescript/frontend/src/app/market/[market]/page.tsx
@@ -37,10 +37,13 @@ interface EmojicoinPageProps {
 const EmojicoinPage = async (params: EmojicoinPageProps) => {
   const { market: marketSlug } = params.params;
   const names = pathToEmojiNames(marketSlug);
+  console.log({names});
   const hex = names
     .map(n => SYMBOL_DATA.byName(n)?.hex.slice(2))
     .join("");
+  console.log({hex});
   const res = await fetchLatestMarketState(`0x${hex}`);
+  console.log({res});
 
   if (res) {
     const marketID = res.marketID.toString();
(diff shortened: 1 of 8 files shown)


alnoki (changes requested):
- [x] Fix broken Vercel deployment
- [x] Fix CI lints

alnoki (changes requested):
- [x] When I click on the market that is highlighted on the home page, it tries to redirect to the market ID per the old schema, and gives "Emoji not found" error. This should be abstracted as much as possible so that it is only updated in one place for both the cards and for the top one
- [x] How about using an underscore instead of `%20` , e.g `Japanese%20"application"%20button` becomes `Japanes_"application"_button`? (I was originally going to suggest a dash `-` but then I checked `hex_codes` and saw some have a dash in their name, but I don't see any with an underscore). ~Alternatively use `+` as this is how Google, for example, denotes a space~
- [x] Using a comma as a delimiter could be a problem, because some supported coin symbol emojis use on. For example `Person: medium-light skin tone, blond hair`. I'm also concerned that a comma might interfere with chained query parameter parsing. How about a plus? e.g. `peach+peach`?  ~Or if that isn't supported maybe a dunder? `peach__peach`~ (Also consider `Flag: Heard & McDonald Islands`, which as an ampersand `&` )

`folded_hands+folded_hands`

CRBl69:
Noting that `;` is used instead of `+` because Vercel converts the `+` to ` ` and cannot differentiated from a real ` `.

alnoki (approved):

PR #162, opened 2024-07-17
[ECO-1972] add search on home page

CRBl69:
# Description

This PR adds the functionality to search for pools on the home page.

In order to make this process more usable, I made the home page hybrid.

It will use the props provided by the server to display the initial page. But
then, instead of refreshing the page whenever new data needs to be fetched, it
will be fetched dynamically.

# Testing

See vercel.

### DIFF
--- a/src/typescript/frontend/src/components/pages/home/components/emoji-table/components/Search.tsx
+++ b/src/typescript/frontend/src/components/pages/home/components/emoji-table/components/Search.tsx
@@ -11,9 +11,10 @@ import Image from "next/image";
 const searchIcon = <Image className="med-pixel-search" alt="search" src={icon} />;
 
 export const SearchComponent = () => {
-  const setMode = useInputStore(state => state.setMode);
+  const setMode = useInputStore((state) => state.setMode);
   useEffect(() => {
     setMode("home");
+    /* eslint-disable-next-line react-hooks/exhaustive-deps */
   }, []);
   return (
     <Flex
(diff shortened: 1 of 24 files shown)


src/typescript/frontend/src/components/pages/home/components/emoji-table/components/Search.tsx
```diff
@@ -1,33 +1,23 @@
...
-
   return (
     <>
       <InputGroup
-        label="Search:"
-        textScale={{ _: "pixelHeading4", laptopL: "pixelHeading3" }}
+        textScale="pixelHeading3"
         variant="fantom"
-        forId="search"
+        width="unset"
```
xbtmatt, line 10: We can probably set a max value for this (if this is the search box I'm thinking of)

And then maybe add a border on the right? Right now the `X` is like unbounded and just kind of in the middle of the search box area, if we add a border to specifically the right side of the actual input search area it will look a bit cleaner imo, even if it doesn't line up with other things.

xbtmatt: Screenshot 2024-07-17 at 11 57 23 PM

specifically talking about this

src/typescript/frontend/src/components/pages/home/components/emoji-table/index.tsx
```diff
@@ -13,27 +15,135 @@ import SearchComponent from "./components/Search";
...
 }
 
+const encodeEmoji = (emojis: string[]) => {
+  const encoder = new TextEncoder();
+  return emojis.reduce(
+    (acc: string, emoji) =>
+      `${acc}${[...encoder.encode(emoji)].map((e) => e.toString(16)).join("")}`,
+    "0x"
+  );
+};
```
xbtmatt, line 37:
```suggestion
const encodeEmojis = (emojis: string[]) => {
  const encoder = new TextEncoder();
  return "0x" + emojis.reduce(
    (acc: string, emoji) =>
      `${acc}${[...encoder.encode(emoji)].map((e) => e.toString(16)).join("")}`,
    "",
  );
};
```

Let's put this in the `sdk` where the other emoji data helpers go. I find myself using this a lot as well, it'd be nice to have this in the `sdk` as a helper function.

Also moved the `0x` solely for readability, even though yours was already correct obviously :)

CRBl69:
- Add curly braces at the end and start of search
- Add go to first and go to last button for pagination

CRBl69:
- Make search icon a little smaller to fit in brackets

PR #166, opened 2024-07-19
[ECO-1977] Update version badge

CRBl69:
# Description

This PR updates the version badge.

There is not enough space to display it at the top on most displays.

Therefore, it is only show in the footer ~and never on mobiles~.

It is show on mobiles in the footer. When there's not enough space, it will be shown under the emojicoin logo instead of besides it.

# Testing

See vercel.

# Checklist


src/typescript/frontend/src/components/footer/index.tsx
```diff
@@ -21,7 +21,7 @@ const Footer: React.FC = () => {
         <StyledContainer>
           <StyledSocialWrapper>
             <StyledClickItem>
-              <LogoIcon width="170px" cursor="pointer" versionBadge={true} />
+              <LogoIcon width="170px" cursor="pointer" versionBadge={true && !isMobile} />
```
xbtmatt, line 24: You could also just change the flex style (from row to col) for mobile so that they're stacked on top of each other and don't add horizontal space

CRBl69: I think that was the issue this PR was supposed to fix in the first place, the fact that they would stack up on mobile.

alnoki: - [ ] If resizing to a small screen, put the alpha badge below, but enforce no line breaking inside the badge

xbtmatt (commented):
I'm trying to remember why we decided not to have it in the header- it looks like it'd fit pretty similarly to the footer as long as you do the same responsive changes as the width decreases.

I.e., is there a specific reason we're not doing the same thing with the header as we are with the footer? I think it looks great in the footer, would be nice to have it in the header, because honestly I don't think anyone is going to see the footer, especially on the main page.

CRBl69:
This was never discussed on call. I thought about doing it when doing it for the footer, but I didn't find that very beautiful. Please let me know if I should do it. It's a quick fix then we can merge.

xbtmatt:
Yes, let's add it in. I think you might just need to add custom media queries for it is all. If not, even easier. But I think it looks good and it looks odd without the alpha now since I'm so used to it :)

xbtmatt (commented):
Awesome, much better for responsiveness.

Is this a design from figma or can we alter it a bit?

If you don't mind I think a smaller border radius and slightly smaller font would be great, something like this:
Screenshot 2024-07-26 at 12 11 31 PM

 - 8px border radius
 - Roughly ~75% of the font size/height
 - A few more pixels of space between the logo and the version indicator

xbtmatt (approved):
Looks amazing!

PR #173, opened 2024-07-23
[ECO-2002] Fix alignment of featured market emoji

CRBl69:
# Description

The emoji of the featured market was a bit off center due to CSS priority of classes.

This PR fixes that.

# Testing

See vercel preview.

# Checklist


xbtmatt (changes requested):
Looks good visually, but let's refactor these classes to not use `!important`. It's more maintainable and readable in the long run.

I think you can do it by splitting up some of the `#styled-emoji` and `#styled-double-emoji` selectors for padding

PR #178, opened 2024-07-25
[ECO-1764] Resolve names using ANS

CRBl69:
# Description

This PR uses ANS to resolve names for addresses and show them wherever addresses show up.

- It will cache up to 3000 names.
- Each name has a max lifetime of a week.

# Testing

See vercel preview.

Notably, chats, trades, and wherever your address shows up like in the header.

# Checklist


src/typescript/frontend/src/components/pages/emojicoin/components/chat/components/message-container/index.tsx
```diff
@@ -44,7 +44,10 @@ const MessageContainer: React.FC<MessageContainerProps> = ({ message }) => {
               href={toExplorerLink({ value: message.version, linkType: "version" })}
             >
               <Text textScale="pixelHeading4" color="lightGray" textTransform="uppercase">
-                {truncateAddress(message.sender)}
+                {/* Note: someone could have a name starting with "0x", is it a problem ? */}
```
xbtmatt, line 47: Good point, I think you could add a flag to the data passed in (`message` prop I think...? Add a `ansName?: boolean` field) to indicate its an ANS name and truncate it based on that

CRBl69: I thought about that, but I wondered if it was useful or not.

Let's say my name is of the form 0x..., I probably want it formatted as an address, right ? Thought on this ?

src/typescript/frontend/src/components/pages/emojicoin/components/trade-history/index.tsx
```diff
@@ -66,6 +69,22 @@ const TradeHistory = (props: TradeHistoryProps) => {
...
+    (async () => {
+      const names = await getANSNames(sortedSwaps.map((e) => e.swapper));
+      setNames(names);
+    })();
+  }, [sortedSwaps]);
+
+  const sortedSwapsWithNames = useMemo(() => {
+    const data = sortedSwaps.map((e) => ({
+      ...e,
+      swapper: names.get(e.swapper) ? names.get(e.swapper) : e.swapper,
```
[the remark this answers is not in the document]
CRBl69: Oh right, it was `swapper: names.get(e.swapper) ? someFn(names.get(e.swapper)) : someOtherFn(e.swapper),` before. Forgot to change form when removing those calls.

src/typescript/frontend/src/components/pages/emojicoin/components/chat/components/message-container/index.tsx
xbtmatt: Yes I was thinking about this as well, I actually think you're right. If it's in the form of an address just truncate it as such.

You can just call `AccountAddress.isValid` with `strict: true` to check if the address is a valid account address with the `ts-sdk` for this

src/typescript/frontend/src/components/pages/emojicoin/components/trade-history/index.tsx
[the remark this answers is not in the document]
xbtmatt: No worries, I figured you just missed it. I know you're pretty familiar with ts/js 😛

src/typescript/frontend/src/utils/ans.ts
```diff
@@ -0,0 +1,105 @@
...
+
+function serializeMap(map: ANSMap) {
+  let array = Array.from(map.entries()).filter((a) => a[1].expiry > new Date().getTime());
+  if (array.length > MAX_ITEMS_IN_CACHE) {
+    array = array
+      .sort((a, b) => a[1].expiry - b[1].expiry)
+      .slice(array.length - MAX_ITEMS_IN_CACHE);
+  }
+  return JSON.stringify(array);
+}
```
xbtmatt, line 24:
```suggestion
type ANSMapValue = { name: string | null; isNull: boolean; expiry: number };
type ANSMap = Map<string, ANSMapValue>;
type ANSPromiseMap = Map<string, Promise<ANSMapValue>>;

function serializeMap(map: ANSMap) {
  const now = new Date().getTime();
  const notExpired = (entry: [string, ANSMapValue]) => entry[1].expiry > now;

  const res = Array.from(map.entries())
    .filter(notExpired)
    .sort((a, b) => a[1].expiry - b[1].expiry)
    .slice(0, MAX_ITEMS_IN_CACHE);
    
  return JSON.stringify(res);
}
```

Can clean this up a little bit, feel free to use this or change it or whatever, but making it a bit more readable is nice

You might need to add a `//prettier-ignore` above the `const res = ...` line. I personally think the multiple chained function calls on a single line is nearly unreadable so I always try to format it as multiple lines

xbtmatt: Also double check my logic on this, free-handed it in the github comment

src/typescript/frontend/src/components/pages/emojicoin/components/chat/ChatBox.tsx
```diff
@@ -118,6 +120,22 @@ const ChatBox = (props: ChatProps) => {
...
+    (async () => {
+      const names = await getANSNames(sortedChats.map((chat) => chat.user));
+      setNames(names);
+    })();
+  }, [sortedChats]);
+
+  const sortedChatsWithNames = useMemo(() => {
+    const data = sortedChats.map((e) => ({
+      ...e,
+      user: names.get(e.user) ? names.get(e.user) : e.user,
```
xbtmatt, line 133:
```suggestion
      user: names.get(e.user) ?? e.user,
```

src/typescript/frontend/src/components/pages/emojicoin/components/chat/ChatBox.tsx
```diff
@@ -118,6 +120,22 @@ const ChatBox = (props: ChatProps) => {
     /* eslint-disable react-hooks/exhaustive-deps */
   }, [props.data.chats.length, chats.length]);
 
+  useEffect(() => {
+    (async () => {
+      const names = await getANSNames(sortedChats.map((chat) => chat.user));
```
xbtmatt, line 125:
```suggestion
      const users = sortedChats.map((chat) => chat.user);
      const names = await getANSNames(users);
```

Another small readability thing, I know it's a bit more verbose but it's just a lot easier to read

src/typescript/frontend/src/components/pages/emojicoin/components/trade-history/index.tsx
```diff
@@ -66,6 +69,22 @@ const TradeHistory = (props: TradeHistoryProps) => {
     /* eslint-disable react-hooks/exhaustive-deps */
   }, [props.data.swaps.length, swaps.length]);
 
+  useEffect(() => {
+    (async () => {
+      const names = await getANSNames(sortedSwaps.map((e) => e.swapper));
```
xbtmatt, line 74:
```suggestion
      const addresses = sortedSwaps.map((e) => e.swapper);
      const names = await getANSNames(addresses);
```

src/typescript/frontend/src/components/pages/emojicoin/components/trade-history/table-row/index.tsx
```diff
@@ -130,7 +130,9 @@ const TableRow = ({ item, showBorder }: TableRowDesktopProps) => {
               " my-auto ml-auto mr-[20px]"
             }
           >
-            {"0x" + truncateAddress(item.swapper).substring(2).toUpperCase()}
+            {item.swapper.startsWith("0x")
+              ? "0x" + truncateAddress(item.swapper).substring(2).toUpperCase()
```
xbtmatt, line 134:
```suggestion
              ? prettifyHex(item.swapper)
```

There's a `prettifyHex` function I added to `frontend/src/lib/utils` a few days ago you can use here

src/typescript/sdk/src/utils/misc.ts
```diff
@@ -171,7 +171,18 @@ export const truncateAddress = (input: HexInput, numChars: number = 4): string =
     s = input;
   }
   const res = normalizeHex(s);
-  return `${res.substring(0, numChars + 2)}...${res.substring(res.length - numChars, res.length)}`;
+  const start = res.substring(0, numChars + 2);
+  const end = res.substring(res.length - numChars, res.length);
+  return `${start}...${end}`;
+};
```
xbtmatt, line 177: ty for cleaning my mess up 😄

xbtmatt (changes requested):
This is a good start, but I think we need to add another layer of state management to separate the data fetching layer from the data storage layer.

Right now the architecture is:

 - Fetch the names from local storage, remotely if they're not in the map
 - Retrieve the map from local storage in `getANSNames`
 - This recreates the entire map every time you call `getANSNames`. This _might_ be fine, but the # of names could potentially be in the thousands, so reading directly from local storage and recreating the map each time a component loads/displays the names might actually result in a small delay
 - The wallet ANS name is fetched separately and uses separate logic, meaning the logic for resolving a non-wallet address/name is different, and each component that displays an address has to to resolve every single address conditionally as such.
 - That is, there isn't a single source of truth for name lookup, so you can't just call `getANSName(someAddress)` without looking for it in multiple places

I think we need to add a state management layer with zustand/immer. I think we can add this to the `WebSocketEventsProvider` (and rename it to `ZustandStateProvider` or something, since it's already merging two zustand `stores` already)

You can use `Map`s in immer pretty intuitively, you can see how I did it with `markets` in `event-store.ts`, you can pretty much just add a `addANSNames` function in the new `NameStore`.

So you should:

1. Create `name-store.ts`, with `zustand` *and* `immer` (because `immer` has better support for `Map` state updates), see `event-store.ts` for how I did this already
2. Add the resulting `NameStore` to `WebSocketsEventsProvider` in `WebSocketContextProvider.tsx`
3. Add the `useNameStore` hook to the `websockets-context/index.tsx` (which you can also rename to be `zustand-context/index.tsx` or something)
4. Expose the map as a variable in a type `NameStore`, aka `names`
5. Add an "action" called `addName` where you will call `set((state) => state.names.set(address, name)` or something like this, again see `event-store.ts` for an example
6. Move the `ansPromiseMap` and integrate it, along with the deduplication logic, into the state store `addName` function
7. Use `localStorage` to initialize the store. i.e., pass in the `localStorage` deserialized map values as the `initialState` to the `createNameStore` function you create
8. Add a `useEffect` to `AptosContextProvider` that gets and/or adds the current wallet address to the name store, which will trigger when the address changes.
9. Add a concept of the `network` to the name lookup map in the `NameStore`, either by prepending it as the value to the map `key` or by creating multiple maps per network. The former is probably easier/simpler overall. Right now all the caching/fetching functionality is unaware of the current network other than being in the fetch call
10. All of the above allows you to seamlessly use the name store in every component where you display an address, simply by calling `const names = useNameStore((s) => s.names);` and then displaying it with just `names.get(user) ?? user`.


There's a few reasons for this:
1. We consolidate the name lookup to a single source of truth- that way you don't have to worry about it in the frontend component. As mentioned in `10.` above, your component need not be aware of the state management/business logic for how the names are resolved. All it does it look it up in the map, even if it's the currently connected user's wallet address.
2. `names` map updating will cause a re-render, so the data will always be up to date
3. You can technically update/display each name as each promise resolves instead of awaiting all of them at once as long as you call `addName` per promise instead of waiting for all of them to finish first.
4. If we decide later to pre-render the top 10 names on a leaderboard or something, we can do it server-side and load that data into the state store without having to worry about coalescing that data other than by simply calling `addName`
5. All business/deduplication/caching code will exist in the fetch functions and the state store instead of in components, making them much cleaner overall
6. The map object isn't fully recreated every time you fetch new names because `zustand/immer` are optimized to handle map value updates without fully recreating objects

I know that's a lot but should be enough to get you through most of it. Let me know if any of it is confusing or there's something missing!

src/typescript/frontend/src/components/pages/emojicoin/components/trade-history/index.tsx
```diff
@@ -66,6 +66,18 @@ const TradeHistory = (props: TradeHistoryProps) => {
...
+
+  const sortedSwapsWithNames = useMemo(() => {
+    const data = sortedSwaps.map((e) => ({
+      ...e,
+      swapper: nameResolver(e.swapper),
+    }));
+    return data;
+    /* eslint-disable react-hooks/exhaustive-deps */
+  }, [nameResolver]);
+
```
xbtmatt, line 80: So I think the logic for this should actually just be in a hook, and then you use the hook anywhere you use a potentially resolved ANS name:

```typescript
// The hook
const useNameResolver = (address: AccountAddressInput) => {
  const normalizedAddress = AccountAddress.from(address).toString();
  const resolvedName = useNameStore((s) => s.names.get(normalizedAddress) ?? normalizedAddress);

  return resolvedName;
}

// ...then, in the body of the displaying component
const name = useNameResolver(someAddress);

// ...
<span>{name}</span>
// ...
```

And it will force a re-render only when the entry in the map changes (I'm pretty sure, at least, haha, might need to test)

This also ensures that you don't have to mutate/get rid of the swapper address in the swap data. You can keep `swapper` as is and then use the hook anywhere you need the name, usually further down the component tree

src/typescript/frontend/src/lib/store/name-store.ts
```diff
@@ -0,0 +1,173 @@
...
+      uncachedNames.push(address);
+    }
+  }
+  if (uncachedNames.length > 0) {
+    for (const address of uncachedNames) {
+      if (!ansPromiseMap.has(address)) {
+        const x = fetchANSName(address)
+          .then((res) => ({
+            name: res,
+            isNull: res === null,
```
xbtmatt, line 141: Don't think you use this, yea? Did you mean to

xbtmatt (commented):
Awesome man this is nice, thank you for doing all this. It makes things a lot easier in the end

The only two things are

1. Put that resolve logic into a hook for easier use
2. Make sure you normalize the addresses when you check for and store them

src/typescript/frontend/src/lib/store/name-store.ts
```diff
@@ -0,0 +1,173 @@
...
+  return createStore<NameStore>()(
+    immer((set, get) => ({
+      ...initialState,
+      getMapWithNames: (addresses) => {
+        return getNameMap(set, get, addresses);
+      },
+      getResolverWithNames: (addresses) => {
+        const map = getNameMap(set, get, addresses);
+        return (address) => map.get(address)?.name ?? address;
+      },
```
xbtmatt, line 170: I think you won't need these anymore if you use the hook instead, btw, all you need really is just the map, which will notify the hook of changes

src/typescript/frontend/src/components/pages/emojicoin/components/trade-history/index.tsx
CRBl69: IDK about this. The reason I did it this way is that I want to be able to pass an array to the initial function that will query the names. This is so that `localStorage` is written to only once. With your implementation, we will write to `localStorage` multiple times, and potentially slow the down the program. I think it's better to leave it this way. We can discuss this on a call.

xbtmatt: Discussed on eng call, agreed with you here, you can keep it as is as an array

xbtmatt (commented):
Looks good to me, just need to resolve the linter issues and the comments I mentioned, I think there's some unused variables and cspell stuff

src/typescript/frontend/src/lib/store/name-store.ts
CRBl69: Yes I do use it, line 153. It is stored in the promise map.

src/typescript/sdk/src/utils/misc.ts
```diff
@@ -171,9 +171,31 @@ export const truncateAddress = (input: HexInput, numChars: number = 4): string =
...
+    return `${start}...${end}`;
+  }
+  return input;
+};
+
+/**
+  * Takes an address or a name and formats it accordingly.
+  */
+export const formatDisplayName = (input: string) => {
+  if (AccountAddress.isValid({input: input, strict: false}).valid) {
```
xbtmatt, line 192:
```suggestion
  if (AccountAddress.isValid({ input, strict: false }).valid) {
```

src/typescript/frontend/src/lib/store/name-store.ts
xbtmatt: `isNull`? I didn't mean the value `x`, just the `isNull` field, sorry

xbtmatt (commented):
Just the few small things

xbtmatt (approved):

src/typescript/frontend/src/lib/store/name-store.ts
xbtmatt: Feel free to resolve, not a big deal either way

PR #181, opened 2024-07-25
[ECO-2011] Add registrant grace period flag

alnoki:
# Changes

## Production code

Store a `RegistrantGracePeriodFlag` under a market object address during market
registration, which ensures that only the market registrant may swap during the
first 5 minutes after market registration.

Anyone else may swap after the registrant has swapped, or after the grace period
has lapsed, whichever comes first.

Note that the actual logic changes are minimal, see `emojicoin_dot_fun.move`.

## Test code

1. Add setup helper function to ensure `AptosCoin` is initialized, to fix tests
   broken by upstream changes to Aptos Framework since the last updates to this
   package.
1. Fix older tests, which were unaware of the new grace period logic and which
   were broken by new `E_NOT_REGISTRANT` assert, through the introduction of
   a test helper function that disables the registrant grace period check.
1. Add tests to verify grace period logic.

## Docs

1. Re-build docs.

# Testing

## e2e tests

e2e tests now fail cc @xbtmatt 

## Publication verifiability

Published on devnet

## Coverage testing for new logic

Previously the package was tested to 100% coverage, but the coverage tool breaks
for inline functions (https://github.com/aptos-labs/aptos-core/issues/9154) so
re-demonstrating a 100% coverage run requires manually taking out `inline`
keywords as per 6d65d23 and c13dbe2.

Instead to verify coverage for the new logic, from `src/move/emojicoin_dot_fun`:

```sh
aptos move test --coverage --dev
aptos move coverage bytecode --dev --module emojicoin_dot_fun
```

Then see references to `RegistrantGracePeriodFlag` under `entry public swap`,
which should be marked green indicating 100% coverage:

```asm
[51]	211: Exists[16](RegistrantGracePeriodFlag)
[51]	212: BrFalse(246)
B25:
[8]	213: CopyLoc[1](market_address: address)
[8]	214: ImmBorrowGlobal[16](RegistrantGracePeriodFlag)
[8]	215: StLoc[98](flag_ref: &RegistrantGracePeriodFlag)
[8]	216: ImmBorrowLoc[90](event: Swap)
[8]	217: ImmBorrowField[49](Swap.time: u64)
[8]	218: ReadRef
[8]	219: CopyLoc[98](flag_ref: &RegistrantGracePeriodFlag)
[8]	220: ImmBorrowField[50](RegistrantGracePeriodFlag.market_registration_time: u64)
[8]	221: ReadRef
[8]	222: LdConst[43](U64: [0, 163, 225, 17, 0, 0, 0, 0])
[8]	223: Add
[8]	224: Gt
[8]	225: StLoc[103](grace_period_over: bool)
[8]	226: CopyLoc[198](swapper_address: address)
[8]	227: MoveLoc[98](flag_ref: &RegistrantGracePeriodFlag)
[8]	228: ImmBorrowField[51](RegistrantGracePeriodFlag.market_registrant: address)
[8]	229: ReadRef
[8]	230: Eq
[8]	231: StLoc[199](swapper_is_registrant: bool)
[8]	232: MoveLoc[103](grace_period_over: bool)
[8]	233: MoveLoc[199](swapper_is_registrant: bool)
[8]	234: Or
[8]	235: BrFalse(237)
B26:
[7]	236: Branch(243)
B27:
[1]	237: MoveLoc[0](swapper: &signer)
[1]	238: Pop
[1]	239: MoveLoc[139](market_ref_mut: &mut Market)
[1]	240: Pop
[1]	241: LdConst[23](U64: [16, 0, 0, 0, 0, 0, 0, 0])
[1]	242: Abort
B28:
[7]	243: CopyLoc[1](market_address: address)
[7]	244: MoveFrom[16](RegistrantGracePeriodFlag)
[7]	245: Pop
```

# Checklist


xbtmatt (commented):
Looks good to me, but what is the reasoning behind ending the grace period early? What do you think about making the grace period the same amount of time no matter what?

Few reasons:
  1. It makes a countdown simpler on the frontend
  2. It creates a feeling of suspense and anticipation 
  3. If the grace period ends after the first registrant trade, the UI will have to instantly update to reflect that. We can do it with mqtt/wss but I worry that the UX for a user waiting for the countdown to end might be a little confused if/when the countdown goes from `4:36` instantly to `0:00`
  4. You have to very clearly warn the registrant that they only get one trade instead of a simple grace period
  5. It also makes it a little easier for us to index/get markets that are registrant-only, since we know it's always X minutes from registration

CRBl69:
I'll note that ending the grace period early will probably make is to that bots have less perfect timing when the grace period ends.

CRBl69 (approved):

xbtmatt (approved):
Reconsidered and think discouraging/mitigating botting is a bigger benefit than UX clarity since we can ultimately come up with a solution for that pretty easily.

Looks good to me!

PR #184, opened 2024-07-28
[ECO-1942] Fix emoji picker bugs; allow mobile users to use native emoji keyboards

xbtmatt:
# Description

- [x] Allow mobile users to use native emoji keyboards
- [x] Filter/sanitize this input
- [x] Fix the issue with the emoji picker automatically showing up when navigating to a new page
- [x] Fix the issue with invisible emojis sometimes (by using `useCallback` instead of `useRef`, essentially we weren't updating the component ref when the DOM reloaded the textarea)
- [x] Fix general UX issues where state was decoupled- we generally always show the picker when the input is focused, otherwise don't show it
- [x] The only exception to this is when the page first loads- we focus the input box but don't show the emoji picker yet.

# Testing

Try it on mobile.

# Checklist


src/typescript/frontend/src/components/pages/emoji-picker/EmojiPicker.tsx
```diff
@@ -198,7 +198,7 @@ export default function EmojiPicker(props: HTMLAttributes<HTMLDivElement>) {
           onClickOutside={onClickOutside}
           perLine={8}
           // TODO: Use this instead of the current "not allowed" emoji we're using..?
-          // exceptEmojis={["🔥", "🚀", "🌙", "🌟", "🎉", "🎊", "🎈", "🎁", "🎆", "🎇"]}
+          exceptEmojis={["🔥", "🚀", "🌙", "🌟", "🎉", "🎊", "🎈", "🎁", "🎆", "🎇"]}
```
xbtmatt, line 224: Didn't mean to add. It w ok nt do anything anyway because it used text IDS, not emojis. Fixed in next PR if not this one

alnoki (approved):

PR #186, opened 2024-07-29
[ECO-2008] Use the new balance variable to calculate user rank

CRBl69:
# Description

This PR uses the new balance field introduced in Inbox to calculate user rank.

As this field is not present in MQTT events, some custom logic has been done to
the event pipeline to accomodate for that.

# Testing

See vercel.

# Checklist


src/typescript/frontend/src/lib/store/event-store.ts
```diff
@@ -305,8 +310,16 @@ export const createEventStore = (initialState: EventState = defaultState) => {
...
+        if (get().guids.has(event.guid) && !event.guid.startsWith("Swap")) return;
+        if (
+          event.guid.startsWith("Swap") &&
+          (event as Types.SwapEvent).balanceAsFractionOfCirculatingSupply
+        ) {
+          updateLocalStorage(event);
+          return;
+        } else {
+          addToLocalStorage(event);
+        }
```
xbtmatt, line 322: Trying to understand what exactly is the logic here; why are you not returning if the state store already has the `Swap` event?

Also I wrote some type guards for this to get the compile time types:

```typescript
const event: AnyEmojicoinEvent = ...; // some swap event data
const { isSell } = event; // Compiler/IDE would yell at you for this
if (isSwapEvent(event)) {
  // It's now typed as a Types.SwapEvent event
  // You can now access swap-only fields
  const { isSell } = event;
}
```

src/typescript/frontend/src/lib/store/event-utils.ts
```diff
@@ -189,11 +189,12 @@ export const getLatestBars = (market: MarketStateValueType) => {
   return RESOLUTIONS_ARRAY.map((res) => ({ ...market[res]!.latestBar, period: res }));
 };
 
+const shouldKeep = (event: AnyEmojicoinEvent) => {
+  const time = getEmojicoinEventTime(event);
+  return time / 1000n > BigInt(new Date().getTime() - 60 * 1_000);
+};
```
xbtmatt, line 195:
```suggestion
// Perhaps put in a `const.ts` file somewhere more appropriate
export const LOCAL_STORAGE_EXPIRY_TIME_MS = 60 * 1000;

const shouldKeep = (event: AnyEmojicoinEvent) => {
  const eventTimeMs = Number(getEmojicoinEventTime(event) / 1000n);
  const now = new Date().getTime();
  return eventTimeMs > now - LOCAL_STORAGE_EXPIRY_TIME_MS;
};
```

I know I'm being a little neurotic about these things so bare with me, but for the sake of readability I think we can split the logic up here

xbtmatt (commented):
Looks simple enough- mostly just curious what the logic behind the local storage stuff is

src/typescript/frontend/src/lib/store/event-store.ts
CRBl69: So this is because now you have two types of Swap events. The ones coming from WebSockets, which are the same as before, and the ones coming from the `inbox_swaps` endpoint, which have the balance field. I do this so that if the event does not have the field, it is replaced by the event with the field present.

xbtmatt: Ahhh okay, gotcha, that makes sense.

src/typescript/frontend/src/lib/store/event-store.ts
```diff
@@ -305,8 +310,16 @@ export const createEventStore = (initialState: EventState = defaultState) => {
       // Note that we `unshift` here because we add the latest event to the front of the array.
       // We also update the latest bar if the incoming event is a swap or periodic state event.
       pushEventFromClient: (event) => {
-        if (get().guids.has(event.guid)) return;
-        addToLocalStorage(event);
+        if (get().guids.has(event.guid) && !event.guid.startsWith("Swap")) return;
+        if (
+          event.guid.startsWith("Swap") &&
+          (event as Types.SwapEvent).balanceAsFractionOfCirculatingSupply
```
xbtmatt, line 316:
```suggestion
          typeof (event as Types.SwapEvent).balanceAsFractionOfCirculatingSupply === "bigint"
```

Zero is falsy here so if it's `0n` it won't enter the truthy half of the if/else block, gotta explicitly check for the type. You could also do `... !== "undefined"`

xbtmatt (approved):
LGTM just make sure to fix the falsy thing, otherwise might get some unxpected behavior there

Feel free to resolve the conversation once you see it

PR #187, opened 2024-07-29
[ECO-1858] Add VPN, geoblocking feature

CRBl69:
# Description

This PR adds geoblocking using Vercel's API, and VPN blocking via VPNAPI.io .

# Testing

See vercel preview.

# Checklist


xbtmatt (commented):
Was confused for a sec then instantly realized that meant it's working, haha!

Screenshot 2024-07-31 at 11 44 04 PM

Also it looks like you included some other commits, you mind rebasing so I can view the files that have actually changed? 😄

src/typescript/frontend/src/utils/geolocation.ts
```diff
@@ -0,0 +1,44 @@
...
+  const cache = localStorage.getItem(COUNTRY_LOCALSTORAGE_KEY);
+
+  if (cache) {
+    const data: LocationCache = JSON.parse(cache);
+    if (data.expiry > new Date().getTime()) {
+      return data.country;
+    }
+  }
+
+  const queryResult = await fetch(`https://ipinfo.io/json?token=${IPINFO_TOKEN}`)
```
xbtmatt, line 26: If we accidentally ddos this or site goes down no one can get into the site, yea? Right now I can only get to the verify screen btw, so no one will be able to see anything if that happens

xbtmatt (commented):
Aside from the comments my main question is is the banner supposed to show up for me, having been geoblocked? I can only see the very dark `Connect Wallet` prompt- obviously can't click it

src/typescript/frontend/src/utils/geolocation.ts
```diff
@@ -0,0 +1,44 @@
+import { LazyPromise } from "@sdk/utils";
+import { GEOBLOCKING_ENABLED } from "lib/env";
+
+const COUNTRY_CACHING_DURATION = 1000 * 60 * 60 * 24; // One day
```
xbtmatt, line 4:
```suggestion
const COUNTRY_CACHING_DURATION = 1000 * 60 * 60 * 24; // One day.
```

src/typescript/frontend/tailwind.config.js
```diff
@@ -74,11 +74,16 @@ module.exports = {
         "0%": { transform: "translateX(0)" },
         "100%": { transform: "translateX(-4527.83px)" },
       },
+      banCarousel: {
+        "0%": { transform: "translateX(0)" },
+        "100%": { transform: "translateX(-5008.25px)" },
```
xbtmatt, line 79: I see you figured out the ways of the infinite carousel too...(this was such a pain lol, had to measure it so it wouldn't snap at the end of the animation)

src/typescript/frontend/src/components/header/wallet-button/ConnectWalletButton.tsx
```diff
@@ -62,39 +66,53 @@ export const ButtonWithConnectWalletFallback: React.FC<ConnectWalletProps> = ({
...
+  const inner =
+    !connected || !children || isBanned ? (
+      <Button
+        className={
+          className + (mobile ? " px-[9px] border-dashed border-b border-b-dark-gray" : "")
+        }
+        disabled={isBanned}
+        onClick={(e) => {
+          e.preventDefault();
+          onClick ? onClick() : openWalletModal();
```
xbtmatt, line 78:
```suggestion
          (onClick ?? openWalletModal)();
```

I think this is my code :P definitely prefer the nullish coalescing operator for readability

src/typescript/frontend/src/components/geoblocking/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <div className="w-full bg-pink">
+      <div className="overflow-hidden w-full flex-row">
+        <div className="flex">
+          <div className="flex gap-[16px] animate-banCarousel">{items}</div>
+        </div>
+      </div>
+    </div>
+  );
+};
```
xbtmatt, line 36:
```suggestion
export const GeoblockedBanner: React.FC = () => {

  return (
    <div className="w-full bg-pink">
      <div className="overflow-hidden w-full flex-row">
        <div className="flex">
          <div className="flex gap-[16px] animate-banCarousel">
            {Array.from({ length: 4 }).map((_, i) => (
              <Item key={i} />
            ))}
          </div>
        </div>
      </div>
    </div>
  );
};
```

Can always do something like this as well, personally I try to avoid creating React/JSX elements in the render function body and prefer to put them in the return function instead, since it makes it a little more clear what the actual DOM structure will be when rendered

But really it's fine either way

src/typescript/frontend/src/hooks/use-is-banned.ts
```diff
@@ -0,0 +1,12 @@
+import { useState } from "react";
+import { isBanned } from "utils/geolocation";
+
+const useIsBanned = () => {
+  const [banned, setBanned] = useState(true);
+
+  isBanned().then((res) => setBanned(res));
+
+  return banned;
+};
```
xbtmatt, line 10: Unfortunately this will call the function dozens of times- you could use `useRef` here to make the value stable across re-renders, but even then it'd still call it every time you use the hook in a different component.

If you had to do it yourself, since it's an app-wide global state check, something like this should really be in its own context provider or in a zustand state store to ensure there's a single state for the promise. Plus, you probably also want to implement locking across multiple components as well

But since this is just a fetch call this is what `@tanstack/react-query` is for really, because it has its own app-wide global cache and key mechanism already 😄 

Or you can do it server side and cache the result. React/nextjs purportedly say multiple server-side async fetches are stable across several components, so you theoretically wouldn't have to care about state

CRBl69: This is not a problem.

Using the `LazyPromise` object, the calculations only happen once.

To make sure of what I'm saying, I added a `console.log` statement in the `getCountry` function, and I only see it printed once.

Yes, `isBanned` might be called multiple times, but here is the code of that function:

```
if (!GEOBLOCKING_ENABLED) return false;
return !isAllowedCountry(await country.get());
```

First, it's checking a constant. Then, it's calling `isAllowedCountry` (which is just a comparison) on `await country.get()`. `country` here is the `LazyPromise`, which basically ensures that the calculation only happens once, and all subsequent requests are served the computed value obtained on the first calculation.

It works. If you're ok with this, I wouldn't change it.

xbtmatt: I think it would mostly *work* as long as it stays exactly the way it is, but if there are any changes to how we manage geoblocking it will almost certainly need to be changed

It's also just generally not a great practice to put something in the function body if the intention isn't to call the function each render cycle. You could achieve the same exact effect you have now with a `useEffect` or even a `useEffect` that stores a boolean ref (to be stable across component mount/unmount).

But in that case again you're asusming the state would never change, and it might, if for example, the fetch call fails and we need to re-fetch

xbtmatt (changes requested):
After thinking on this a bit more I think we have to consider the extent to which we ensure the user is not geoblocked. I don't know if we can say we made a good faith effort to block users if we just store their country in local storage, since that doesn't even require a VPN

But also, once you actually do store the country (with a VPN for example), it's in there forever and then they don't even have to use the VPN anymore.

I think this should *at minimum* be designed similarly to how the `middleware` for the galxe campaign works, where we store a hashed value in the user's cookies (you could just store this in local storage too) that represent proof of having been authenticated by the server at some point.

But also we want to ensure that users visiting are authenticated every time unlike the wallet, because the entire intention is to not let geoblocked users use the app. If we only check it once and never again, we can't actually say we blocked access to the site based on geolocation to the best of our knowledge.

Part of this depends on how stringent we need to be with geoblocking but I assume that's the entire point, so I think this probably needs to be refactored to use server-side authentication with middleware and most likely no caching or localstorage

xbtmatt (commented):
Ok so I think we mainly just need to fix some of the React stuff- the geoblocking mechanism itself is fine, I thought about it a bit more and it doesn't need to be super overkill as long as the cache is refreshing every week or whatever

But for what it's worth- since we're using headers, we could also just add this to middleware and show the user a different server component based on the result of their location without having to use any client-side code to determine it.

But what we have here is fine!

xbtmatt (changes requested):
I think we need to refactor this to use cookies and achieve the same effect, just server-side.

You can either await the promise server-side and render the app conditionally based on the awaited value (is or is not banned)

But if your worry is making the website take longer to load because of this then we can pass the promise from the server to the app's outermost containers and wrap them in <Suspense ... /> while the promise resolves, meaning they'll display the loading fallback until it's resolved.

If we don't do either of these though, and it's all client-side, then you'll see the banner flash from `BANNED` to the real value every single time you open the web app, which we definitely don't want.

src/typescript/frontend/src/utils/geolocation.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export type Location = {
+  country: string;
+  region: string;
+  countryCode: string;
+  regionCode: string;
+  vpn: boolean;
+};
+
+const isAllowedLocation = (location: Location) => {
```
xbtmatt, line 11:
```suggestion
const isNotAllowedLocation = (location: Location) => {

// or maybe

const isLocationGeoblocked
```

Let's make the phrasing of the boolean geoblocked logic functions to be consistent so that it's more natural to read, otherwise it gets a little confusing reading it.

I think this and `isBanned` are the only two we use. Also let's change `isBanned` to `isUserGeoblocked` just so to keep all the language consistent

xbtmatt (commented):
Awesome, looks much more straightforward with the server component passing the geoblocked status down as a prop! :D

The only thing is I'm on the preview page and it looks like the env vars are set, but I'm not geoblocked..?

xbtmatt (changes requested):
Overall looks great- the banner is nice :)

But I did set my cookies manually to test what the site is like without the galxe verification, and the problem is if the user has verified somehow already and they're banned, they can still do everything on the site. You can connect your wallet from anywhere on the site in most places that allow for interaction, like the swap button for example

The best way to block a user from actually using the site is by intercepting everything that goes through the wallet provider.

Basically you should pass the Aptos context provider (in `providers.tsx` the `geoblocked` server component prop, and if they're blocked, just do nothing when they try to submit a transaction.

I think there's only two functions that are used for submission and they both consolidate to a single function- you can just do a check for `geoblocked` in that function and return early if so

BTW if you want to avoid passing the prop around the whole app you can just store it in state somewhere, that way you can avoid prop drilling and passing it around everywhere

It's a quick fix though, shouldn't be very complex

alnoki (dismissed):
- [x] Fix `pre-commit`
- [x] Document new environment variables and their types in example env file

xbtmatt (commented):
Just commenting here that the main two things left are:

1. Adjust the wording of `isBanned` to be consistent with the rest of the wording (`isAllowedCountry` or something)
2. Ensure that the site properly blocks geoblocked users from using anything related to the wallet adapter for the site (insert a check in the adapter to disallow them submitting anything if they're geoblocked)

xbtmatt (approved):
Awesome, none of the stuff works! :D Thanks for all the changes on this, it's working great

PR #189, opened 2024-07-30
[ECO-2043] Add grid market reorder functionality; various miscellaneous updates

xbtmatt:
# Description

- [x] Add reorder animations to the home page emoji grid when bump order is selected for the sort and the user is on page 1
- [x] Add reorder animations to the trade history entries
- [x] Add reorder animations to the chat history
- [x] Debounce the animation *as well as* put it on an interval
- [x] Stop glow animations from happening when the component is unmounted and when the event didn't happen recently (when the user switches filters and the grid re-orders, for example)

Note a lot of these weren't originally in the scope of this PR, but working with the data was impossible with some of the current functionality.

- [x] Add mqtt connection management (leaving this alone after this since probably we're not going to use it soon)
  - [x] Added an automatic unsubscription interval that checks if the markets are active or ont
  - [x] Automatically unsubscribes after the component unmounts 
- [x] Cleaned up `useEffect`s in the codebase to avoid leaks
- [x] Removed the aggregate market fetch for the home page
  - [x] Left it in for the chart because it needs to be able to see it in order to show data for the market. Otherwise initially it will show the ghost. I remembered that that is the reason I did it initially
- [x] Add the animation toggle in. It only disables grid animations for now, but it's easy to use it later, since we store it in the user settings provider, see below
- [x] Fix the market registration animation continuing even after the user navigates away
- [x] Fix the issue where the market registration animation plays if you go back to the page within a few seconds
- [x] Squished the hero image down a bit so it doesn't affect the layout- aka capped it at 60% of the viewport
- [x] Cleaned up the animations for the grid size changing. It now fades in and out when the number of elements per row changes to avoid layout snapping/thrashing.

```typescript
  const animate = useUserSettings((s) => s.animate);
  // And this is used for the switcher.
  const toggleAnimate = useUserSettings((s) => s.toggleAnimate);
```

- [x] Refactored the search page to entirely use search components to avoid double fetching and incorrect fetching
- [x] Resolved UX issues with the picker and search bar on home page
- [x] Keep the featured market static based on highest sorted daily volume

# Testing

Incredible amounts of visual testing.

# Checklist


src/typescript/frontend/src/components/pages/home/components/table-card/animation-variants.ts
```diff
@@ -0,0 +1,60 @@
+const DURATION = 0.3;
+const INSERTION_DELAY = 0.2;
+const PER_ELEMENT_DELAY = 0.02;
+
+// This isn't the longest animation ("initial") is- but for the purpose of updating the grid
```
xbtmatt, line 5:
```suggestion
// This isn't the longest animation ("initial" is), but for the purpose of updating the grid
```

alnoki (approved):

PR #193, opened 2024-08-05
[ECO-2055] Add playwright with minimal test

CRBl69:
# Description

This PR adds:

- Playwright.
- Example e2e caching test.
- `/test` route wihch returns the latest transaction version.
- GitHub action for Playwright.

# Testing

Clone this branch and run `pnpm run e2e`.

# Checklist


src/typescript/frontend/e2e/example.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { test, expect } from '@playwright/test';
+
+const sleep = (ms: number) => {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
```
xbtmatt, line 5: nit, but there's a sleep function in the `sdk/utils` already btw

src/typescript/frontend/src/app/test/route.ts
```diff
@@ -0,0 +1,10 @@
+import { getAptos } from "lib/utils/aptos-client";
+import { NextResponse } from "next/server";
+
+export const revalidate = 2;
```
xbtmatt, line 4: Not a big deal with just a single test but because we'll be using varying cache times in different tests you'll probably have to abstract some of the data passed in to them to a function, and you don't want to pass magic numbers to those functions

Ideally we export the expected test's revalidation time as a constant somewhere and then get it from there, just like how we do `REVALIDATION_TIME`

src/typescript/frontend/e2e/example.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+
+test('api test route', async ({ page }) => {
+  const numeicRegex = /^[0-9]+$/;
```
xbtmatt, line 8:
```suggestion
  const numericRegex = /^[0-9]+$/;
```

xbtmatt (commented):
Think we need to add a `prepare` script to set up the repo, or at least `pnpm exec playwright install` prepended to the pnpm `e2e` test script command

Also did you actually get this to pass the tests? It's not caching anything for me unless I visit the page manually (which is fine- as long as we just understand the behaviour and write tests around it)

xbtmatt (commented):
Looks like the yaml parser is really mad at you 😛

Even without the tests working yet, the explanation on cache hits/misses are really nice, really appreciate the detail on that, that helps a *lot*! It looks good so far, just request review again after you get the tests working in CI and whatnot

xbtmatt (approved):
Awesome! Thanks for doing this one too, lots of nice things to have

The only thing that would be nice to have in this that I just remembered is a redirect on the test page for when in prod, that way there's no `/test` route on our production page. I think this should work, just place it near the bottom of `next.config.mjs` and test it to make sure

```typescript
const redirectTestPageInProd = {
  redirects: async () => [
    {
      source: ROUTES.test,
      destination: "/404",
      permanent: true,
    },
  ],
};

/** @type {import('next').NextConfig} */
const nextConfig = {
  ...(process.env.NODE_ENV === "production" ? redirectTestPageInProd : {}),
  // Rest of the config...  
```

Other than that looks good!

CRBl69:
Hey. So the caching behavior is not the same in prod and dev modes. For this reason, I run playwright tests against the prod mode. If I hide `/test` in prod mode, the tests won't work anymore.

PR #194, opened 2024-08-08
[ECO-1842] Implement functionality for the provide liquidity button

CRBl69:
# Description

This PR adds the functionality to go to the liquidity page of a market on the market page.

# Testing

See vercel preview.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/components/pages/emojicoin/components/trade-emojicoin/LiquidityButton.tsx
+++ b/src/typescript/frontend/src/components/pages/emojicoin/components/trade-emojicoin/LiquidityButton.tsx
@@ -5,6 +5,7 @@ import Button from "components/button";
 import { translationFunction } from "context/language-context";
 import { isInBondingCurve } from "utils/bonding-curve";
 import { AnimatedProgressBar } from "./AnimatedProgressBar";
+import { Link } from "components";
 
 export const LiquidityButton = (props: GridProps) => {
   const { t } = translationFunction();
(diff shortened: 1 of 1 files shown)


src/typescript/frontend/src/components/pages/emojicoin/components/trade-emojicoin/LiquidityButton.tsx
```diff
@@ -14,7 +15,9 @@ export const LiquidityButton = (props: GridProps) => {
       {!isInBondingCurve(props.data) ? (
         <StyledContentHeader>
           <Flex width="100%" justifyContent="center">
-            <Button scale="lg">{t("Provide liquidity")}</Button>
+            <Link href={`/pools?pool=${props.data.emojis.reduce((p, c) => `${p}${c.emoji}`, "")}`}>
```
xbtmatt, line 18:
```suggestion
            <Link href={{
              pathname: ROUTES.pools,
              query: { pool: props.data.emojis.join("") },
            }}>
```

BTW yesterday I realized we can just use the default export `import Link from "next/link";`to have access to a better Link component with a clearer prop for pathname and query.

The original one you're using wrapped it with theme and styling and wasn't made when we were using `nextjs`, so you can use it here since there's no styling here (although you could add the scramble I suppose)

It's not really a big deal whether or not we use it here so I'll approve, just in case you feel like changing it 😁

xbtmatt (approved):

src/typescript/frontend/src/components/pages/emojicoin/components/trade-emojicoin/LiquidityButton.tsx
CRBl69: My bad. I meant to import the `nextjs` one. Auto-import messed this one up :sweat_smile:

xbtmatt: Hahaha it's ok, we have too many old components lying around. I do it all the time too

xbtmatt (approved):
Awesome, very clean! Looks good to me

PR #199, opened 2024-08-16
[ECO-2076] Incorporate new indexer

CRBl69:
# Description

This PR adds a comprehensive indexer framework:

- Docker compose
- Binaries for processor and broker
- GitHub actions to push images to Docker Hub

Containers are built and pushed when a tag of the form `processor-v*` or `broker-v*` is pushed.

PRs that merged into this branch:
- #207 
- #206 

# Testing

Follow README inside `src/rust`.

# Checklist


cfg/yamllint-config.yaml
alnoki, line 1: As discussed on eng call:

- [ ] Revert changes to config
- [ ] Manually escape on individual files

.github/workflows/push-broker.yaml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: 'ubuntu-latest'
+    steps:
+    - uses: 'actions/checkout@v4'
+    - id: 'metadata'
+      uses: 'docker/metadata-action@v5'
+      with:
+        images: 'econialabs/emojicoin-broker'
+        tags: |
+          type=ref,event=branch
+          type=semver,pattern={{version}}
```
alnoki, line 13: As I recall this won't work becayse `type=semver` requires a tag of semver form, not `emojicoin-indexer-broker-<semver>`

- [ ] Update to replicate `cloud-infra` example

.github/workflows/push-broker.yaml
```diff
@@ -0,0 +1,37 @@
...
+        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
+        push: 'true'
+        tags: '${{ steps.metadata.outputs.tags }}'
+name: 'Build broker Docker image and push to Docker Hub'
+'on':
+  push:
+    branches:
+    - 'unstable'
+    tags:
+    - 'v*'
```
alnoki, line 36: Should specify broker in tag

.github/workflows/push-broker.yaml
```diff
@@ -0,0 +1,37 @@
...
+        file: 'src/indexer/broker/Dockerfile'
+        labels: '${{ steps.metadata.outputs.labels }}'
+        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
+        push: 'true'
+        tags: '${{ steps.metadata.outputs.tags }}'
+name: 'Build broker Docker image and push to Docker Hub'
+'on':
+  push:
+    branches:
+    - 'unstable'
```
alnoki, line 34: I think we might want to just incorporate this as `main` later, to reduce PRs

.github/workflows/push-processor.yaml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: 'ubuntu-latest'
+    steps:
+    - uses: 'actions/checkout@v4'
+    - id: 'metadata'
+      uses: 'docker/metadata-action@v5'
+      with:
+        images: 'econialabs/emojicoin-processor'
+        tags: |
+          type=ref,event=branch
+          type=semver,pattern={{version}}
```
alnoki, line 13: Same as for the broker image build

.github/workflows/push-processor.yaml
```diff
@@ -0,0 +1,37 @@
...
+        file: 'src/indexer/processor/rust/Dockerfile'
+        labels: '${{ steps.metadata.outputs.labels }}'
+        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
+        push: 'true'
+        tags: '${{ steps.metadata.outputs.tags }}'
+name: 'Build processor Docker image and push to Docker Hub'
+'on':
+  push:
+    branches:
+    - 'unstable'
```
alnoki, line 34: Same as for the broker image build

src/indexer/broker/src/server/sse.rs
```diff
@@ -0,0 +1,59 @@
+use std::{convert::Infallible, sync::Arc};
```
alnoki, line 1: This is useful as an example, but how about removing it from getting compiled?

src/indexer/broker/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+mod types;
+mod util;
+
+#[tokio::main]
+async fn main() -> Result<(), ()> {
+    env_logger::init();
+
+    let processor_url = std::env::var("PROCESSOR_WS_URL").unwrap();
+
+    let (tx, _) = broadcast::channel(2048);
```
alnoki, line 14: How about making this an env var?

Just in case AWS wants something special

src/indexer/broker/src/server.rs
```diff
@@ -0,0 +1,54 @@
...
+
+async fn health() {}
+
+pub async fn server(tx: Sender<EmojicoinDbEvent>) -> Result<(), std::io::Error> {
+    let app_state = AppState { tx };
+
+    let app = prepare_app(Router::new().route("/", get(health)));
+    let app = app.with_state(Arc::new(app_state));
+
+    let listener = tokio::net::TcpListener::bind("0.0.0.0:3009").await.unwrap();
```
alnoki, line 48: How about making this port and env var in case AWS wants something?

src/indexer/broker/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    libssl-dev \
+    libpq-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder $BIN_PATH $APP_DIR/
+
+# This helps the container stop faster
+STOPSIGNAL SIGKILL
```
alnoki, line 34: Does the binary also handle `SIGTERM` as stipulated by AWS?

I know docker will send it when you run compose locally, but the binary should also handle it for when AWS sends it in 

Otherwise container updates could hang

src/indexer/broker/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ARG APP_DIR
+ARG RUST_ROOT
+COPY $RUST_ROOT $APP_DIR
+WORKDIR $APP_DIR
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    libudev-dev \
+    build-essential \
+    libclang-dev \
+    libdw-dev \
+    libpq-dev \
```
alnoki, line 15: Why does the broker have a postgres dependency?

src/indexer/broker/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    && rm -rf /var/lib/apt/lists/*
+RUN cargo build --all-features --release -p broker
+
+FROM debian:bookworm-slim
+ARG APP_DIR
+ARG BIN_PATH
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    libssl-dev \
+    libpq-dev \
```
alnoki, line 28: Why does the broker have a postgres dependency?

src/indexer/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
alnoki, line 1: If we combine to one workspace, this isn't needed

src/indexer/Cargo.lock
```diff
@@ -0,0 +1,4542 @@
+# This file is automatically @generated by Cargo.
```
alnoki, line 1: If we combine to one workspace this isn't needed

src/indexer/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[workspace]
+exclude = ["processor"]
```
alnoki, line 2: Why?

src/indexer/compose.yaml
```diff
@@ -0,0 +1,75 @@
...
+    image: 'econialabs/emojicoin-broker'
+    ports:
+    - '3009:3009'
+    restart: 'unless-stopped'
+
+  postgres:
+    environment:
+      POSTGRES_USER: 'indexer'
+      POSTGRES_DB: 'indexer'
+      POSTGRES_PASSWORD: 'indexer'
```
alnoki, line 22: - [ ] Change this and all affected dependents to `emojicoin`, since AWS requires 8 chars or more

src/indexer/compose.yaml
```diff
@@ -0,0 +1,75 @@
...
+    ports:
+    - '5432:5432'
+
+  postgrest:
+    depends_on:
+      postgres:
+        condition: 'service_healthy'
+    environment:
+      PGRST_DB_URI: 'postgres://indexer:indexer@postgres:5432/indexer'
+      PGRST_DB_ANON_ROLE: 'web_anon'
```
alnoki, line 43: Is there a migration that adds this role?

src/indexer/compose.yaml
```diff
@@ -0,0 +1,75 @@
...
+      STARTING_VERSION: '${STARTING_VERSION}'
+      EMOJICOIN_MODULE_ADDRESS: '${CONTRACT_ADDRESS}'
+    depends_on:
+      postgres:
+        condition: 'service_healthy'
+    image: 'econialabs/emojicoin-processor'
+    restart: 'unless-stopped'
+    stop_signal: 'SIGKILL'
+    ports:
+    - '3008:3008'
```
alnoki, line 70: Env var in case AWS has issues

src/indexer/compose.yaml
```diff
@@ -0,0 +1,75 @@
...
+      dockerfile: 'broker/Dockerfile'
+    depends_on:
+      processor:
+        condition: 'service_started'
+    environment:
+      PROCESSOR_WS_URL: 'ws://processor:3008/ws'
+      RUST_LOG: 'info'
+    image: 'econialabs/emojicoin-broker'
+    ports:
+    - '3009:3009'
```
alnoki, line 15: Env var in case AWS has issues

alnoki: For port 3008/3009

src/indexer/compose.yaml
```diff
@@ -0,0 +1,75 @@
...
+      GRPC_AUTH_TOKEN: '${GRPC_AUTH_TOKEN}'
+      GRPC_DATA_SERVICE_URL: '${GRPC_DATA_SERVICE_URL}'
+      STARTING_VERSION: '${STARTING_VERSION}'
+      EMOJICOIN_MODULE_ADDRESS: '${CONTRACT_ADDRESS}'
+    depends_on:
+      postgres:
+        condition: 'service_healthy'
+    image: 'econialabs/emojicoin-processor'
+    restart: 'unless-stopped'
+    stop_signal: 'SIGKILL'
```
alnoki, line 68: AWS is asking for SIGTERM

src/Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
+exclude = ["indexer"]
```
alnoki, line 2: Why? This is complicating the repo layout

.github/workflows/push-broker.yaml
CRBl69: I'm not sure. A tag is for a version that contains a working broker, processor, api and frontend. If we start tagging each component individually, how would we deploy on AWS ? It's easier to say depoly `v1.5.2` rather than deploy `broker-v1.2.3` and `processor-v2.1.0` and `frontend-v1.4.2` and then realize that those are actually not compatible and you need some other version. It's like a DSS version: we don't have a version for the aggregator, one for the processor, one for MQTT, etc., we have a version of the stack. Here it's the same thing IMO.

src/indexer/broker/src/main.rs
CRBl69: This is just the size of the buffer of the broadcast channel (i.e. it will retain up to 2048 unread messages). It has little to do with AWS. I can put it as a `const` so that we don't have magic numbers hanging around.

src/indexer/Cargo.toml
CRBl69: The processor is already in a workspace inside the aptos directory. I gives an error if we add it to the workspace. Can't wait to try the SDK though

src/indexer/broker/src/server/sse.rs
CRBl69: This is already the case. Check `server.rs`. I use feature flags to compile this or not. I just configured docker to not compile this by default.

.github/workflows/push-broker.yaml
alnoki: > If we start tagging each component individually, how would we deploy on AWS?

The stack deployment file simply specifies the image tag for the binaries inside, like a cargo crate specifies versions of each internal dependency. Segmenting versions by service will also help with faster build times and debugging updates to individual services: https://github.com/econia-labs/emojicoin-dot-fun/blob/5557744537021842147eff0be955448c6580e0de/src/indexer/cloud-formation/deploy-dev.yaml#L6

> we have a version of the stack

Yes, this is fine, and will work with the above: once something gets merged to `main` (the testing branch) and works, we merge `main` to `indexer-stable` and issue a tag to it just like on the DSS

src/indexer/compose.yaml
alnoki:
```suggestion
    - '${PROCESSOR_WS_PORT}:${PROCESSOR_WS_PORT}'
```

src/indexer/compose.yaml
alnoki:
```suggestion
```suggestion
    - '${PROCESSOR_WS_PORT}:${PROCESSOR_WS_PORT}'
```
```

alnoki: Addressing in ECO-2115

src/indexer/compose.yaml
alnoki: Addressing in ECO-2115

src/indexer/broker/Dockerfile
alnoki: Addressing in ECO-2115

src/indexer/broker/src/main.rs
alnoki: Addressing in ECO-2115

src/indexer/broker/Dockerfile
alnoki: Addressing in ECO-2115

.github/workflows/push-processor.yaml
alnoki: Addressing in ECO-2115

.github/workflows/push-broker.yaml
alnoki: Addressing in ECO-2115

alnoki (approved):

xbtmatt (approved):

.github/workflows/push-broker.yaml
CRBl69: I think specifying the version in a file is not a good idea. I think it is simpler to have a global version.

AWS would have two deployments:

- The latest global tag.
- Whatever is currently on main.

This way, we can easily track our stable deployment and our edge deployment.

How would we do this with versioning inside a file ?

Also, if we ever need to roll back, it will be close to impossible to know which version of the processor works with which version of the broker.

Since these services are closely linked together and evolve together, it only makes sense to have one version, IMO.

PR #215, opened 2024-08-26
[ECO-2135] Add logging, extend `cargo` `pre-commit` hooks

CRBl69:
# Description

This PR adds:

- logging for the broker

# Testing

Run pre-commit, two `clippy` and `cargo-check` checks should be present.

Run the broker with different log levels (see [env_logger documentation](https://docs.rs/env_logger/latest/env_logger/) on how to do that) and check output.

# Checklist


src/rust/broker/src/server.rs
```diff
@@ -48,6 +51,20 @@ pub async fn server(tx: Sender<EmojicoinDbEvent>, port: u16) -> Result<(), std::
...
+    if cfg!(all(feature = "ws", not(feature = "sse"))) {
+        info!("Starting web server with ws.");
+    }
+    if cfg!(all(not(feature = "ws"), feature = "sse")) {
+        info!("Starting web server with sse.");
+    }
+    if cfg!(all(not(feature = "ws"), not(feature = "sse"))) {
+        warn!("Starting web server with no endpoints.");
+    }
+
```
xbtmatt, line 67: Can we put this in a match statement or is it not possible because it's a macro

src/rust/broker/src/server/ws.rs
```diff
@@ -34,61 +35,56 @@ async fn handle_websocket(socket: WebSocket, state: Arc<AppState>) {
...
 
     tokio::select! {
-        _ = t => {
-            log::warn!("Connection broken (t).")
-        }
-        _ = r => {
-            log::warn!("Connection broken (r).")
-        }
+        _ = t => {}
+        _ = r => {}
```
xbtmatt, line 86: Did we remove this because the error logs are handled elsewhere in every case now?

CRBl69: Yes. Each case has its own internal logging now.

src/rust/broker/src/server.rs
CRBl69: I don't think this is possible. Each on the `cfg!(...)` will resolve to either `true` or `false` at compile time.

CRBl69:
Closing because commits were already merged into main by another PR

PR #220, opened 2024-08-27
[ECO-2142] Implement retry functionality for the broker

CRBl69:
# Description

This PR adds retry functionality for the broker.

A connection is considered successful when a network connection was established and the network connection lasted for at least 10 seconds.

- When starting the broker, if a successful connection cannot be established, exit.
- When a connection breaks, the broker will try to reconnect 10 times.
- When a connection breaks, retry instantly (after 1 second).
- When a retry fails, retry after 15 seconds.
- When no retries are left, exit.
- When a retry results in a successful connection, reset the number of retries left.

# Testing

1. Export `RUST_LOG=broker=trace` to get as much info as possible.
1. Start the broker by itself, assert that it does not try to reconnect.
1. Start the processor with docker compose.
1. Start the broker, it should connect.
1. Wait 10 seconds so that the broker considers the connection successful.
1. Stop the processor.
1. Witness the retries.
1. Start the processor.
1. Witness the successful connection.

# Checklist

### DIFF
--- a/src/rust/broker/src/main.rs
+++ b/src/rust/broker/src/main.rs
@@ -1,5 +1,8 @@
+use std::sync::Arc;
+
 use log::{error, info};
-use tokio::sync::broadcast;
+use serde::{Deserialize, Serialize};
+use tokio::sync::{broadcast, RwLock};
 
 mod processor_connection;
 mod server;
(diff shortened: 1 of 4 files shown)


src/rust/broker/src/processor_connection.rs
```diff
@@ -1,26 +1,126 @@
...
+
+    loop {
+        // Timestamp of the start of the connection.
+        let start_timestamp = SystemTime::now();
+        let res =
+            processor_connection(processor_url.clone(), tx.clone(), broker_health.clone()).await;
+        let connection_duration = SystemTime::elapsed(&start_timestamp).unwrap();
+
+        // In the broker connected to the processor and the connection duration was longer than the
+        // retry treshold, it is considered successful.
```
xbtmatt, line 55:
```suggestion
        // If the broker connected to the processor and the connection duration was longer than the
        // retry threshold, it is considered successful.
```

src/rust/broker/src/main.rs
alnoki, line 1: Use a `live` and `ready` endpoint like PostgREST: https://postgrest.org/en/stable/references/admin_server.html#admin-server

Container self-checks on `live` and load balancer health checks on `ready`

Just return 200 or fail

PR #228, opened 2024-08-31
[ECO-2158] Create a docker image for starting the `emojicoin_dot_fun` local testnet for e2e testing locally and in CI

xbtmatt:
# Description

Since we've now merged [#ECO-2159](https://github.com/econia-labs/cloud-infra/pull/8) and built an `aptos-cli` image from that, we can use it to push a more specific `emojicoin_dot_fun` Aptos node image to Dockerhub

- [x] Complete [#ECO-2159](https://github.com/econia-labs/cloud-infra/pull/8)
- [x] Use the resulting image tag `econialabs/aptos-cli:4.0.0` in a new image that creates the node and any extra data/state we want for e2e/ci tests
- [x] Ensure it works locally
- [x] Ensure it works in CI
  - https://github.com/xbtmatt/emojicoin-dot-fun/actions/runs/10756427484/job/29829272735
  - https://hub.docker.com/repository/docker/xbtmatt/emojicoin-dot-fun-localnet-publisher/tags
- [x] Add the ability to:
  - [x] Restart from genesis
  - [x] Start from the latest version (don't restart at all)
- [x] Fix the issues with restarting, ensure there are no issues with the processor
- [x] Ideally make a simple test in CI to ensure the image actually works
- [x] Update the image to be named `...local-testnet-deployer` or something, not `...aptos-node` (went with `deployer`)

# Testing

  - https://github.com/xbtmatt/emojicoin-dot-fun/actions/runs/10756427484/job/29829272735
  - https://hub.docker.com/repository/docker/xbtmatt/emojicoin-dot-fun-localnet-publisher/tags

# Checklist


src/docker/aptos-node/Dockerfile
alnoki, line 1: I appreciate the abstraction here, you're clearly catching on to the elegance and power of Docker. Since we will re-use this, I think it actually best belongs in the `cloud-infra` repo: if we need a v5.0.0 CLI build in the future, we wouldn't want to have to come back to this repo and bump it.

I just made `econialabs/aptos-cli` on Docker Hub, so you should be able to push there using a new PR on the `cloud-infra` repo:

- [ ] Make `GIT_TAG` into a build-time argument, so that you can pass to docker the version of the CLI you want built
- [ ] Configure an action that extracts the version from the tag and passes it at buildtime (you can use something like the `'${{ steps.metadata.outputs.labels }}'` value that is already in the existing actions, and pass it to the build args option of the `build-push-action` step)
- [ ] Keep the runtime dependencies required, as I think you'll need at least `libssl-dev`, though `git` and `curl` you can probably get rid of

Once that PR merges:
- [ ] Force push the `aptos-node-v4.0.0` tag to `main` so we have an on-main tag, which will just overwrite the image on Docker Hub (I assume you'll off-main tags during PR development to verify the build)
- [ ] Update this image to use `econialabs/aptos-node:4.0.0` as the base image inside of which you publish the package, etc.

alnoki (changes requested):

src/docker/aptos-node/Dockerfile
xbtmatt: Awesome! Yeah I figured it made more sense to have the tag match the CLI version. Makes it easier to manage and understand which version it is

And awesome, so essentially I'll move this current image of just the cli to `cloud-infra` and then use it to build the more specific image which has the emojicoin contract published

xbtmatt: Done!

src/docker/emojicoin-dot-fun-aptos-node/sh/create-publish-payloads.sh
```diff
@@ -0,0 +1,38 @@
...
+	--override-size-check \
+	--included-artifacts none \
+	--package-dir $move_dir/emojicoin_dot_fun/ \
+	--json-output-file $emojicoin_json_path &
+
+aptos move build-publish-payload \
+	--assume-yes \
+	--private-key $PUBLISHER_PK \
+	--encoding hex \
+	--named-addresses rewards=$address,integrator=$address,emojicoin_dot_fun=$address \
```
alnoki, line 34:
```suggestion
	--named-addresses \
            rewards=$address,integrator=$address,emojicoin_dot_fun=$address \
```

.github/workflows/push-aptos-node.yaml
```diff
@@ -1,16 +1,34 @@
...
 jobs:
   build-push:
     runs-on: 'ubuntu-latest'
     steps:
     - uses: 'actions/checkout@v4'
-    - run: 'git submodule update --init --recursive src/rust/processor'
+      with:
+        submodules: 'false'
+    - name: 'Install the latest Aptos CLI'
+      uses: './.github/actions/get-latest-aptos-cli'
```
alnoki, line 19:
```suggestion
      uses: 'aptos-labs/aptos-core/.github/actions/get-latest-cli@8792eef'
```

.github/actions/get-latest-aptos-cli/action.yaml
alnoki, line 1: - [ ] Delete this file in lieu of direct source to reference

.github/workflows/push-aptos-node.yaml
```diff
@@ -19,18 +37,27 @@ jobs:
...
         context: '.'
-        file: 'src/docker/aptos-node/Dockerfile'
+        file: 'src/docker/emojicoin-dot-fun-aptos-node/Dockerfile'
         labels: '${{ steps.metadata.outputs.labels }}'
-        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
+        # The emojicoin-node image can only be built for the platform on which
+        # the build is running. This is because the build process requires
+        # running `aptos node run-localnet`, which spawns a new container. This
+        # new container must be able to run the `aptos` binary, which is only
+        # available for the platform on which the build is running.
```
alnoki, line 52: Functionally I totally understand that in practice this just wasn't working on GitHub actions. Theoretically, as to

> only available for the platform on which the build is running

I suspect that something could be done about this unless you can point me to a reference in docs that explicitly say "you can't do x", but definitely not necessary right now

Because as I understand it, basically everything is cached except maybe the state generation, which someone can always run once if they need to do local testing, and which should take less than a minute

Feel free to resolve this comment unless there is something you disagree on

.github/workflows/push-aptos-node.yaml
```diff
@@ -1,16 +1,34 @@
+# cspell:word localnet
 ---
+env:
+  # yamllint disable rule:key-ordering
+  DOCKER_IMAGE_PLATFORM: 'linux/amd64'
+  BIG_MONEY_GUY_PK: >-
+    0xa5e11183d0c23b8c49b40dd71bba7fa0992aed9b89c42934a1bb8f6e4ccc4ec4
+  PUBLISHER_PK: >-
```
alnoki, line 8: Just noting that for future use we might want to use `SECRET` and `PUBKEY` (both 6 letters, nice for indentation), and no ambiguity because technically private key and public key have the same acronym

Not asking to change, feel free to resolve unless you disagree

.github/workflows/ts-run-tests.yaml
```diff
@@ -10,17 +10,13 @@ jobs:
...
-    - run: |
-        git submodule update --init --recursive src/inbox
-        cp src/inbox/example.env src/inbox/.env && \
-        docker buildx create \
-        --name=mybuilder \
-        --use \
-        --driver=docker-container
-      shell: 'bash'
+    - name: 'Install the latest Aptos CLI'
+      uses: './.github/actions/get-latest-aptos-cli'
```
alnoki, line 17:
```suggestion
      uses: 'aptos-labs/aptos-core/.github/actions/get-latest-cli@8792eef'
```

alnoki (approved):
Feel free to resolve all suggestions upon acceptance and other comments as I flagged, then merge once you've verified it all still works

.github/workflows/push-aptos-node.yaml
xbtmatt: I can use `SK` instead, that would probably be more clear than `PK`. But I generally avoid `public key` since each account's public key is not the same as its address and can have a fairly complex meaning on Aptos (multi-sig keys, multied25519 keys, scp256k1 keys, etc)

alnoki: Agreed that Aptos' schema confuses things even more, what with authentication keys holding more precedence than public keys. Mainly it's just that anywhere keypairs are involved this gotcha seems to keep coming up so I figured I'd flag here. Feel free to resolve

xbtmatt: Resolved in a separate PR to make sure it doesn't break other things: https://github.com/econia-labs/emojicoin-dot-fun/pull/235

.github/workflows/push-aptos-node.yaml
xbtmatt: I may not have written that as clear as I could have; it's not so much "you can't do this in Docker" but more that we'd have to find a solution for getting the container created from `aptos node run-localnet` to match the architecture of the host machine rather than the Docker image that runs the `aptos node run-localnet` command.

The docker startup script in the `run-localnet` command pulls the image at runtime, I assume based on the `DOCKER_DEFAULT_PLATFORM` env var, see [here](https://github.com/aptos-labs/aptos-core/blob/f39917abb975872c447c4aeb4001e9c0248ec13d/crates/aptos/src/node/local_testnet/indexer_api.rs#L108) where Daniel warns about setting it.

Since `Illegal instruction` is explicitly the error code for an incompatible opcode based on the running machine's processor architecture and that's the error I was running into, I assume that's the reason the CLI was not really intended to be run inside a container.

We could:
  1. apply a patch to the CLI docker startup scripts
  2. Try using `DOCKER_DEFAULT_PLATFORM`

But I actually am not sure we should have (or even need) pre-generated state inside a container anymore, for a few reasons that we can discuss later on a call (it will take a bit to explain). The jist of it though is that it's more difficult than I realized and I'm not sure it makes sense anyway

In terms of orchestration we can always just wait on the readiness endpoint aka `http://localhost:8070` to be up, and then go from there.

The main problem that we were running into was more of a devex thing where we weren't properly cleaning up the containers in relation to our indexer processors, but it doesn't need docker orchestration to achieve that- it just needs to be approached holistically instead of piecemeal

src/docker/compose.local.yaml
```diff
@@ -1,51 +1,63 @@
...
-      dockerfile: 'src/docker/aptos-node/Dockerfile'
+    image: 'econialabs/aptos-cli:4.1.0'
+    profiles: ["test-harness"]
     ports:
     - '50051:50051'
     - '8070:8070'
     - '8090:8090'
     - '8080:8080'
     - '8081:8081'
     healthcheck:
```
CRBl69, line 15: You don't need this. It actually gives me a warning when running `docker compose up`. When `network_mode` is set to `host`, this is ignored.

src/docker/compose.local.yaml
```diff
@@ -1,51 +1,63 @@
...
-      interval: '30s'
-      timeout: '10s'
-      retries: 5
-      start_period: '30s'
-    command: ['bash', '/app/run-fullnode.sh']
+      test: ['CMD', 'curl', 'http://localhost:8070/']
+      interval: '5s'
+      timeout: '5s'
+      retries: 10
+      start_period: '60s'
```
CRBl69, line 20: This should go inside the docker image itself (using [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck)). Imagine we use this in another repo, we'll have to copy pasta this there. Then imagine we want to change the `start_period` or `retries`, we'll have to make a PR into every repo where this is used and hope we didn't miss any.

src/docker/compose.local.yaml
```diff
@@ -1,51 +1,63 @@
...
+      args:
+        PUBLISHER_PRIVATE_KEY: '${PUBLISHER_PRIVATE_KEY}'
+        EMOJICOIN_MODULE_ADDRESS: '${EMOJICOIN_MODULE_ADDRESS}'
+    env_file:
+    - path: './example.local.env'
+      required: true
+    - path: './.env'
+      required: false
+    image: 'econialabs/emojicoin-dot-fun-local-deployer:latest'
+    profiles: ["test-harness"]
```
CRBl69, line 54: This is a very interesting setting that I didn't know about. How about merging the two compose files and using profiles instead ? Deploying the indexing stack and pointing it to testnet would look something like `docker compose --profile testnet up -d` and running locally would look like `docker compose --profile local up -d`.

src/sh/emojicoin/wait-for-localnet.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+check_endpoint() {
+	curl -s -f -o /dev/null http://localhost:8070/
```
CRBl69, line 4: See notes about checking with `jq`.

CRBl69 (changes requested):
Some more notes:

I see you build the contract outside docker. Is that truly necessary ? If not, I'd rather have it build inside the docker container. The reason is that this will make it easier IMO to build the docker containers in a more environment agnostic way. Any machine will be able to do that, you don't need `aptos` to be installed on your machine. You will also be able to run docker compose up without having to run preliminary steps.

src/docker/compose.local.yaml
alnoki: > This should go inside the docker image itself 

I would actually advise against health check internalization, because on AWS we configure the health checks externally like this.

src/docker/compose.local.yaml
alnoki: +1 ^

src/docker/compose.local.yaml
CRBl69: Also, I don't think we would deploy the aptos node

src/sh/emojicoin/wait-for-localnet.sh
xbtmatt: Discussed on call - check that the `not_ready` length of the healthcheck endpoint with `jq` has a length of 0:

Something like:

```shell
# Credit to Bogdan :p
jq '.notReady | length'
```

src/docker/compose.local.yaml
xbtmatt: Thanks for noting, was wondering why that warning kept coming up, I'll remove it

src/sh/emojicoin/wait-for-localnet.sh
xbtmatt: Fixed and added to the base image in `cloud-infra`: https://github.com/econia-labs/cloud-infra/pull/11

.github/workflows/push-aptos-node.yaml
alnoki: > I actually am not sure we should have (or even need) pre-generated state inside a container anymore

Agreed, it's only a few commands to publish and fund, and the amount of time it takes is negligible compared to how long it takes the site to boot/run other e2e commands

Probably cleanest to not have anything extra in it

Feel free to mark as resolved

src/docker/compose.local.yaml
alnoki: My mistake, AWS considerations don't apply here (at least for now). Consider resolved on my end, I yield to @CRBl69 for final resolution

src/docker/compose.local.yaml
xbtmatt: They're nice to use in terms of excluding/including specific services, but actually orchestrating the order of services is a bit of a pain...I used a `frontend` profile because it's really the easiest to manage, but using more than that didn't really make sense.

It's all in the `README.md` now, too!

src/docker/compose.local.yaml
xbtmatt: Awesome, Bogdan and I discussed it on call on Friday, decided to keep it in to be explicit, but it's also in the base image as well. (As well as the updated healthcheck)

src/docker/compose.local.yaml
```diff
@@ -1,63 +1,48 @@
...
-    - '8081:8081'
+    container_name: 'localnet'
+    build:
+      context: '../../'
+      dockerfile: 'src/docker/localnet/Dockerfile'
     healthcheck:
-      test: ['CMD', 'curl', 'http://localhost:8070/']
+      test:
+      - 'CMD'
+      - 'sh/healthcheck.sh'
```
CRBl69, line 22: Is there a reason this is here instead of the container itself ?

src/docker/compose.local.yaml
```diff
@@ -1,63 +1,48 @@
...
-    # be mounted to the container. This mount enables the container to start a
-    # new container. Note that this is *not* 'Docker-in-Docker', but rather
-    # 'Docker-outside-Docker'. That is, the container can act as the host in
-    # terms of creating and orchestrating other containers but still appears as
-    # a sibling to the existing containers and any new containers it creates.
+      retries: 5
+      start_period: '30s'
+    entrypoint:
+    - '/bin/bash'
+    - '/app/sh/run-localnet.sh'
```
CRBl69, line 29: I'd remove this as well.

Since this is a core part of the container and how it functions, it only makes sense to put it there IMO.

Also, if we were to change it, we only need to change it in one place.

CRBl69 (approved):
Since these are minor changes only, feel free to update and merge.

If you do not agree with the changes, please refrain from merging and let's discuss on the ENG call.

PR #229, opened 2024-09-01
[ECO-2152] Segment REST and WS endpoints for dual load balancer schema

alnoki:
# Description

This PR segments the REST and WS endpoints into separate subdomains through
internal load balancer refactors that will enable rate limiting and caching in
subsequent updates.

## CloudFormation

1. Connect PostgREST to a private NLB, accessed by new REST API gateway over VPC
   Link, to enable add-on rate limiting and caching in subsequent updates.
1. Provide a custom domain for WS and REST endpoints, leaving broker connected
   to public ALB due to aforementioned difficulties with header forwarding. Note
   that rate limiting can still be applied to the ALB, even though caching is
   not meaningful for the WS endpoint.
1. Update processor health checks per
   https://github.com/econia-labs/aptos-indexer-processors/pull/45.
1. Update broker health checks per #220 and per WS path updates in this PR.
1. Update README with new domain layout, diagnostics commands.

## Broker path updates

1. Update broker WS server to be at `/` on broker.
1. Update broker `/` health check to be `/live`.
1. Bump version for each time that I vacillated about path strategy, for
   consistency with Docker images pushed to Docker Hub.
1. Pin `apt-get install` versions major.
1. Update broker README for new path strategy.
1. Update Docker compose for new path strategy.

# Testing

- [x] I deployed the changes in this PR on a CloudFormation stack then verified
  the README instructions for both public endpoints and bastion host, using the
  CLI to send in chat events.
- [x] I fully de-provisioned the stack by setting `MaybeDeployStack` to `false` to
  verify no update hangs during de-provisioning.
- [x] I then set `MaybeDeployStack` to `true` to verify full stack provisioning
  from genesis, which will happen on `main` when this lands since `main` is
  currently de-provisioned.

# Checklist


CRBl69 (approved):

src/cloud-formation/deploy-dev.yaml
```diff
@@ -1,20 +1,24 @@
...
+  MaybeDeployPostgrest: 'true'
+  MaybeDeployProcessor: 'true'
+  MaybeDeployRestApi: 'true'
+  MaybeDeployRestApiDnsRecord: 'true'
+  MaybeDeployRouteTables: 'true'
+  MaybeDeployStack: 'true'
+  MaybeDeployVpc: 'true'
   Network: 'testnet'
-  ProcessorImageVersion: '0.4.0'
+  ProcessorImageVersion: '0.5.0'
```
xbtmatt, line 21: Are these versions actually changing or are you bumping it to trigger a re-build? Just curious

xbtmatt (commented):
Awesome, although I'm trying to make sure I understand how this reconciles with the previous auto-deployment of `git` updates.

I see that you essentially namespace all of our deployments as subdomains per the root environment subdomain name we provide. Was it handled in `deploy-dev.yaml` with the `Environment: 'dev'` config option?

Also, any way we can change that to `testnet`? Haha just to be a little more explicit about what/where it is- unless I've misunderstood the intention, which I very well may have...

PR #233, opened 2024-09-04
[ECO-2164] Add indexer WAF, REST API caching/auth key

alnoki:
# Description

1. Add header-based REST API key authentication.
1. Add REST API caching.
1. Update README with REST API key instructions.
1. Add a web application firewall (WAF) with assorted managed policies.
1. Add custom WAF rules for REST and WS endpoints, but leave disabled.

# Testing

To verify changes in this PR:
1. Follow the README for the `ECO-2164` stack to see `curl` results with and
   without header.
1. See the Web ACL traffic overview dashboards as described in the README.

# Checklist


xbtmatt (commented):
Awesome, although I'm trying to make sure I understand how this reconciles with the previous auto-deployment of `git` updates.

I see that you essentially namespace all of our deployments as subdomains per the root environment subdomain name we provide. Was it handled in `deploy-dev.yaml` with the `Environment: 'dev'` config option?

Also, any way we can change that to `testnet`? Just to be a little more explicit about what/where it is

src/cloud-formation/indexer.cfn.yaml
alnoki, line 1: @xbtmatt I'm replying to https://github.com/econia-labs/emojicoin-dot-fun/pull/233#pullrequestreview-2281605384 here so it can be contained in a thread

Currently, the domain namespacing is set up around two inputs:

1. The root domain that someone registers on Route 53 (for example `foo.bar`)
1. The `Environment` parameter (for example `dev` in `deploy-dev.yaml` on `main`)

These are smashed together to make <Environment>.<Root domain>, for example `dev.foo.bar` and `ws.foo.bar`

If it were to be additionally labeled with the `Network` parameter, then I think we would still want the environment to be included to prevent namespace collisions if there are two deployments on the same network. This would yield `<Environment>.<Network>.<Root domain>`, for example:

prod.mainnet.foo.bar

fallback.mainnet.foo.bar

Is this what you are asking for?

src/cloud-formation/deploy-dev.yaml
```diff
@@ -1,7 +1,10 @@
 ---
 parameters:
   BrokerImageVersion: '0.7.0'
-  Environment: 'dev'
+  EnableWafRulesGeneral: 'true'
+  EnableWafRulesRestApi: 'false'
+  EnableWafRulesWebSocket: 'false'
+  Environment: 'ECO-2164'
```
alnoki, line 7: - [ ] Revert to `dev` after final review

CRBl69 (approved):

PR #234, opened 2024-09-04
[ECO-2104] Update market registration components

CRBl69:
# Description

This PR updates the market registration components:

- Add disclaimer about the 4 APT deposit.
- Add disclaimer about the grace period.
- Precisely calculate and display the cost of the transaction.

# Testing

Register a market on vercel.

# Checklist


src/typescript/frontend/src/lib/hooks/queries/use-num-markets.ts
```diff
@@ -0,0 +1,20 @@
...
+  return RegistryView.view({ aptos }).then((res) => Number(res.n_markets));
+}
+
+export const useNumMarkets = () => {
+  const res = useQuery({
+    queryKey: ["num-markets"],
+    queryFn: () => {
+      return getNumMarkets();
+    },
+    staleTime: 30000,
```
xbtmatt, line 16: I think you can make this shorter, like 10-15 seconds, since it's from the REST API and it can handle it

You could also add the length of the market `symbols` map in `useEventStore` in the `queryKey`, which would trigger a re-fetch whenever it changes. You might be able to use a more appropriate way to do this with `tanstack/react-query`, but I think this would work:

```typescript
export const useNumMarkets = () => {
  const numMarkets = useEventStore((s) => s.symbols.size); // or whatever it is, I think it's a set so it's `size`
  // ...
    queryKey: ["num-markets", numMarkets],
  // ...
}
```

I believe that would work by basically triggering a re-fetch whenever the event store sees a new market come in. Then you get the best of both worlds, the node API + the live registrations coming in

src/typescript/sdk/src/const.ts
```diff
@@ -58,6 +58,10 @@ export const BASE_VIRTUAL_CEILING = 49_000_000_000_000_000n;
 export const QUOTE_VIRTUAL_CEILING = 1_400_000_000_000n;
 export const POOL_FEE_RATE_BPS = 25;
 export const MARKET_REGISTRATION_FEE = 100_000_000n;
+export const MARKET_REGISTRATION_DEPOSIT = 400_000_000n;
```
xbtmatt, line 61:
```suggestion
export const MARKET_REGISTRATION_DEPOSIT = 4 * ONE_APT_BIGINT;
```
Maybe a bit clearer this way

xbtmatt (commented):
Looks good so far, I wonder how all those formatting changes snuck in...think we need to consolidate our `.eslintrc.js` files or something

src/typescript/frontend/src/lib/hooks/queries/use-num-markets.ts
CRBl69: Thank you. I tried doing this with `tanstack/react-query` but I couldn't make it work, I'll add this !

src/typescript/sdk/src/emojicoin_dot_fun/emojicoin-dot-fun.ts
```diff
@@ -292,6 +293,42 @@ export class RegisterMarket extends EntryFunctionPayloadBuilder {
...
+  static async getGasCost(args: {
+    aptosConfig: AptosConfig;
+    registrant: AccountAddressInput; // &signer
+    registrantPubKey: PublicKey;
+    emojis: Array<HexInput>; // vector<vector<u8>>
+  }): Promise<{ data: { amount: number; unitPrice: number }; error: boolean }> {
+    const { aptosConfig } = args;
+    const payloadBuilder = new this({
+      ...args,
+      integrator: "0x0000000000000000000000000000000000000000000000000000000000000001",
```
xbtmatt, line 305:
```suggestion
      integrator: AccountAddress.ONE,
```

I believe this should work, a bit more readable

xbtmatt (approved):
Awesome, thanks for the changes- looks good to me! Just change the long `0x0000..1` to `AccountAddress.ONE` and it's good to go

PR #237, opened 2024-09-05
[ECO-2161] Update market page with new grace period dynamics

CRBl69:
# Description

This PR integrates the new grace period dynamics:

- When the market is in a grace period:
    - the *BONDING CURVE: XY%* is replaced with *GRACE PERIOD (X MINUTES AND Y SECONDS LEFT)*
    - if the current user is not the market registrant:
        - the swap button is disabled
        - the swap button has an on hover popup that explains why it is disabled

In order to prevent the market registrant to trade on his own market, in case
of query failure, the UI will let anyone try to trade.

On trade, the grace period is automatically updated !

# Screenshots

image

# Testing

Go to vercel, register a market, go on the market page with two different accounts (registrant and another one).

# Checklist

### DIFF
--- a/src/typescript/frontend/src/lib/hooks/queries/use-grace-period.ts
+++ b/src/typescript/frontend/src/lib/hooks/queries/use-grace-period.ts
@@ -1,18 +1,115 @@
-import { getRegistrationGracePeriodFlag } from "@sdk/markets";
+import { getRegistrationGracePeriodFlag } from "@sdk/markets/utils";
+import { standardizeAddress } from "@sdk/utils/account-address";
 import { useQuery } from "@tanstack/react-query";
+import { useEventStore } from "context/event-store-context";
 import { useAptos } from "context/wallet-context/AptosContextProvider";
+import { useEffect, useMemo, useState } from "react";
 
-export const useGracePeriod = (symbol: string) => {
+// -------------------------------------------------------------------------------------------------
+//
+//                        Utilities for calculating the number of seconds left.
+//
+// -------------------------------------------------------------------------------------------------
+const nowSeconds = () => Math.floor(new Date().getTime() / 1000);
+
+const calculateSecondsLeft = (marketRegistrationTime: bigint) => {
+  const registrationTime = marketRegistrationTime;
+  const asSeconds = Math.floor(Number(registrationTime / 1_000_000n));
+  const plusFiveMinutes = asSeconds + 300;
+  return Math.max(plusFiveMinutes - nowSeconds(), 0);
+};
+
+const formattedTimeLeft = (secondsRemaining: number) => {
+  const remainder = secondsRemaining % 60;
+  const minutes = Math.floor(secondsRemaining / 60);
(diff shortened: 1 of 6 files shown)


src/typescript/frontend/src/lib/hooks/queries/use-grace-period.ts
```diff
@@ -0,0 +1,115 @@
+import { getRegistrationGracePeriodFlag } from "@sdk/markets/utils";
+import { standardizeAddress } from "@sdk/utils/account-address";
+import { useQuery } from "@tanstack/react-query";
+import { useEventStore } from "context/event-store-context";
+import { useAptos } from "context/wallet-context/AptosContextProvider";
+import { useEffect, useMemo, useState } from "react";
+
+// -------------------------------------------------------------------------------------------------
+//
+//             Utilities for calculating the number of seconds left.
```
xbtmatt, line 10:
```suggestion
//                    Utilities for calculating the number of seconds left.
```

src/typescript/frontend/src/lib/hooks/queries/use-grace-period.ts
```diff
@@ -0,0 +1,115 @@
+import { getRegistrationGracePeriodFlag } from "@sdk/markets/utils";
+import { standardizeAddress } from "@sdk/utils/account-address";
+import { useQuery } from "@tanstack/react-query";
+import { useEventStore } from "context/event-store-context";
+import { useAptos } from "context/wallet-context/AptosContextProvider";
+import { useEffect, useMemo, useState } from "react";
+
+// -------------------------------------------------------------------------------------------------
+//
+//                    Utilities for calculating the number of seconds left.
```
xbtmatt, line 10:
```suggestion
//                      Utilities for calculating the number of seconds left.
```

src/typescript/frontend/src/lib/hooks/queries/use-grace-period.ts
```diff
@@ -0,0 +1,115 @@
+import { getRegistrationGracePeriodFlag } from "@sdk/markets/utils";
+import { standardizeAddress } from "@sdk/utils/account-address";
+import { useQuery } from "@tanstack/react-query";
+import { useEventStore } from "context/event-store-context";
+import { useAptos } from "context/wallet-context/AptosContextProvider";
+import { useEffect, useMemo, useState } from "react";
+
+// -------------------------------------------------------------------------------------------------
+//
+//                      Utilities for calculating the number of seconds left.
```
xbtmatt, line 10:
```suggestion
//                        Utilities for calculating the number of seconds left.
```

xbtmatt (approved):
Awesome, I sort of re-arranged a few things, but mostly kept the logic the same.

The main difference is I put the refetching logic into `useQuery` in the `enabled` field and then stop fetching when the market is out of the grace period.

I also changed the format to be `00:00` for the time left, since it was too big when looking at it on mobile. That was it though :D

PR #238, opened 2024-09-06
[ECO-2169] Fix animations by updating the component `keys` for the emoji grid items

xbtmatt:
# Description

The animations got screwed up a bit back when I was fixing the lack of a grid line when search emojis are used.

I added the `index` to each grid item component key, meaning it re-animates any time its position in the grid changes, but that's not what we want, because for bump order in particular, the animations are set up to be triggered by changes to the `orderedGrid` variable, not the component key. (They're set up this way by `framer-motion`, which animates the grid order, indicated by the presence of the `layout` prop in an animated component)

In order to fix this bug while still respecting the original fix, I've added the `searchEmojis` to the component keys and removed the `index` from the component key.

Thus the `searchEmojis` changing will trigger a re-render, which subsequently forces the component to re-checks if the grid line on its left should be present. However, if the grid re-orders due to the `bumpOrder` changing, the component will animate properly because its key will remain the same, since the `index` is no longer in the key. 

# Testing

Tested it on the `production` branch to visually ensure things are working as expected now, but the fixes can go in right now to `main` and will work as expected down the road when everything is set up

# Checklist

### DIFF
--- a/src/typescript/frontend/src/components/pages/home/components/emoji-table/AnimatedClientGrid.tsx
+++ b/src/typescript/frontend/src/components/pages/home/components/emoji-table/AnimatedClientGrid.tsx
@@ -161,7 +161,7 @@ export const LiveClientGrid = ({
       {ordered.map((v) => {
         return (
           <TableCard
-            key={`live-${v.marketID}${v.searchEmojisKey}`}
+            key={`live-${v.marketID}-${v.searchEmojisKey}`}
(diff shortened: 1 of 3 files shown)


CRBl69 (approved):
:+1:

alnoki (approved):

CRBl69 (approved):

PR #242, opened 2024-09-08
[ECO-2177] Convert SDK and e2e tests to use the new `emojicoin` indexer processor

xbtmatt:
# Description

It was a nightmare to rebase #201 , so this PR exists. A more descriptive description will be added later.

- [x] Fix CI/e2e tests by adding `--override-size-check` to the package publication command
- [x] Add new `swap` entry function argument
- [x] Add new `provide_liquidity` entry function argument
- [x] Add new `remove_liquidity` entry function argument
- [x] Fix fund amounts for market registration to account for the initial 4 coin deposit
- [x] Rename `ONE_APTN` to `ONE_APT_BIGINT`
- [x] Change `MARKET_REGISTRATION_FEE` to `APT_BALANCE_REQUIRED_TO_REGISTER_MARKET`
- [x] Update the `APT_BALANCE_REQUIRED_TO_REGISTER_MARKET` to be `5n * ONE_APT_BIGINT` instead of  `1n * ONE_APT_BIGINT`

# Testing

That's entirely what this PR is for. CI tests will be run.

# Checklist


src/docker/compose.local.yaml
```diff
@@ -40,7 +40,7 @@ services:
         PUBLISHER_PRIVATE_KEY: '${PUBLISHER_PRIVATE_KEY}'
     environment:
       PUBLISHER_PRIVATE_KEY: '${PUBLISHER_PRIVATE_KEY}'
-    image: 'econialabs/emojicoin-dot-fun-deployer:latest'
+    image: 'xbtmatt/emojicoin-dot-fun-deployer:0.0.1'
```
xbtmatt, line 43: Update this when the image is available before merging

src/typescript/sdk/src/indexer-v2/queries/base.ts
```diff
@@ -0,0 +1,112 @@
...
+    .eq("market_id", marketID)
+    .eq("period", period)
+    .gte("start_time", startTime)
+    .order("start_time", ORDER_BY.ASC);
+
+const selectMarketLatestStateEvents = ({ marketID }: { marketID?: AnyNumberString }) =>
+  (typeof marketID === "undefined"
+    ? postgrest.from(TableName.MarketLatestStateEvent).select("*")
+    : postgrest.from(TableName.MarketLatestStateEvent).select("*").eq("market_id", marketID)
+  ).order("market_nonce", ORDER_BY.DESC);
```
CRBl69, line 46: There should ideally be an index for this.

I guess this (the order by) is to get the latest updated markets right ?

I think it will mostly be ok, but if we get over 10K markets we might run into slow queries without an index.

(This is a note, feel free to mark as resolved once read.)

src/typescript/sdk/src/indexer-v2/queries/base.ts
```diff
@@ -0,0 +1,112 @@
...
+
+const selectUniqueMarkets = () =>
+  postgrest.from(TableName.MarketLatestStateEvent).select("market_id");
+
+const selectUserLiquidityPools = ({ provider }: { provider: Account }) =>
+  postgrest
+    .from(TableName.UserLiquidityPools)
+    .select("*")
+    .eq("provider", toAccountAddressString(provider))
+    .order("market_nonce", ORDER_BY.DESC);
```
CRBl69, line 62: Not sure if this order by is useful here. This would be used on the pools page, and the only sortings we user there are APR, TVL, 24H volume and all time volume.

We might need an SQL function for this that basically joins `user_liquidity_pools` and the markets table for a given address so that the result returned is only the pools that the user has participated in, and then we'll be able to sort by the previously mentioned characteristics on that.

src/typescript/frontend/src/components/pages/emojicoin/components/trade-history/index.tsx
```diff
@@ -15,12 +15,12 @@ import "./trade-history.css";
...
 
 const toTableItem = (value: Types.SwapEvent): TableRowDesktopProps["item"] => ({
-  ...getRankFromSwapEvent((value.balanceAsFractionOfCirculatingSupply ?? 0) * 100),
+  ...getRankFromEvent(value),
   apt: value.quoteVolume.toString(),
   emoji: value.baseVolume.toString(),
   date: new Date(Number(value.time / 1000n)),
   type: value.isSell ? "sell" : "buy",
-  price: value.avgExecutionPrice.toString(),
+  price: value.avgExecutionPriceQ64.toString(),
```
CRBl69, line 23: Shouldn't this be converted to non-q64, or did I miss the conversion somewhere ?

src/typescript/sdk/src/indexer-v2/queries/base.ts
xbtmatt: 1. Essentially we need to join the user liquidity pools  with the pool data
2. Indicate in `user_liquidity_pools` whether or not the user has any lp coins- check the writeset for the balance of `lp_coin_amount` at the end of the transaction
3. Consider storing quote/lp_coin_amount in the join to indicate whether or not a user has coins

src/typescript/frontend/src/components/pages/emojicoin/components/trade-history/index.tsx
xbtmatt: Nice catch, will convert it at this point (when calculating `price`)

xbtmatt: Fixed in 2d50728

src/typescript/sdk/src/indexer-v2/queries/base.ts
xbtmatt: Discussed on call, this will be in another PR- the index will now be on market nonce and not bump time!

CRBl69 (approved):

alnoki (approved):

CRBl69 (approved):

PR #250, opened 2024-09-16
[ECO-2185] Fix emoji input bug on mobile and registration bug

CRBl69:
# Description

This PR fixes two problems:

- first, the emoji input did not work on mobile. I fixed that.
- second, the market registration gas prediction was too low. Maybe because it was done for the new contract and I was testing with the old contract. I increased it so that it can work with both.


src/typescript/sdk/src/const.ts
```diff
@@ -59,7 +59,7 @@ export const QUOTE_VIRTUAL_CEILING = 1_400_000_000_000n;
 export const POOL_FEE_RATE_BPS = 25;
 export const MARKET_REGISTRATION_FEE = ONE_APTN;
 export const MARKET_REGISTRATION_DEPOSIT = 4n * ONE_APTN;
-export const MARKET_REGISTRATION_GAS_ESTIMATION_NOT_FIRST = 6000;
+export const MARKET_REGISTRATION_GAS_ESTIMATION_NOT_FIRST = ONE_APT * 0.005;
```
xbtmatt, line 62: Does it really cost this much? I think you can use 300,000 as a safe estimate (0.003). 200,000 is the default for most things now from what I remember, so ideally we'd just use that unless you found that it results in it failing (I think you mentioned this, not sure)

xbtmatt (commented):
Just the two small things, we can discuss tomorrow or you can respond here, most likely a quick merge after that

src/typescript/frontend/src/components/pages/launch-emojicoin/hooks/use-register-market.ts
```diff
@@ -33,9 +33,20 @@ export const useRegisterMarket = () => {
...
+        publicKey = new Ed25519PublicKey(
+          typeof account!.publicKey === "string" ? account!.publicKey : account!.publicKey[0]
+        );
+      } catch (_) {
+        return {
+          error: true,
+          data: {
+            amount: 0,
+            unitPrice: 0,
+          }
```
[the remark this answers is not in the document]
CRBl69: Yes. It seems that Aptos Connect public keys are not ed25519 public keys... I was unable to find reliable information on how to handle those.

src/typescript/sdk/src/const.ts
CRBl69: image


I put this just to make sure.

Maybe we can remove after we do not use the old contract anymore. But for now, while I'm still working with the old one for these PRs, I need this.

src/typescript/frontend/src/components/pages/launch-emojicoin/hooks/use-register-market.ts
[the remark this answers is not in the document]
xbtmatt: Commenting here to remind you that you can try a few different things:

1. You can try using a placeholder public key, since I believe that simulations don't verify that the public key matches the authentication key. So it can possibly be anything, you will need to test this
2. To find the public key type, try to create it with different public key types with the SDK and then match the resulting account address to the public key. I don't think this is 100% reliable if they have rotated their key. You could always do a check for this (although I think this is excessively robust/unnecessary)

src/typescript/sdk/src/const.ts
xbtmatt: Gotcha, does the new one cost less gas? Thanks for making sure this works, appreciate the attention to detail here

src/typescript/frontend/src/components/emoji-picker/EmojiPickerWithInput.tsx
alnoki, line 1: Feel free to resolve once all Linear checkboxes checked

src/typescript/sdk/src/const.ts
CRBl69: I think the new one uses less yes

xbtmatt (approved):
Test on mobile and make sure it works, then we can merge this

xbtmatt (approved):
Awesome, just rebased and tested on mobile locally and on the preview build, works great in both cases!

PR #256, opened 2024-09-20
[ECO-2024] Add price feed

CRBl69:
# Description

Add price feed.

# Testing

See vercel.

# Checklist


src/typescript/frontend/src/hooks/use-node-dimensions.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    observer.observe(element);
+
+    return () => {
+      observer.disconnect();
+    };
+  }, [ref]);
+
+  return windowDimensions;
+}
```
xbtmatt, line 39:
```suggestion
```

```react
  const [offsetHeight, setOffsetHeight] = useState(0);

  const elementRef = useCallback((node: HTMLElement | null) => {
    if (node) {
      setOffsetHeight(node.offsetHeight);
    }
  }, []);
  
  return (
    <div ref={elementRef}>
      {offsetHeight}
    </div>
  );
```


You can actually use a neat-o little trick here, where you pass a `useCallback` function to the `ref` prop. You get a stable reference to the element that also re-renders when the element ref changes (meaning your ref is always current)

I used one in the header component at one point, you can see here:

https://github.com/econia-labs/emojicoin-dot-fun/blob/210f9e8082b02fb55090ed9986443e7b6e2563c6/src/typescript/frontend/src/components/header/index.tsx#L33

src/typescript/frontend/src/hooks/use-window-dimensions.ts
```diff
@@ -0,0 +1,24 @@
...
+    function handleResize() {
+      setWindowDimensions(getWindowDimensions());
+    }
+
+    window.addEventListener("resize", handleResize);
+    return () => window.removeEventListener("resize", handleResize);
+  }, []);
+
+  return windowDimensions;
+}
```
xbtmatt, line 24:
```suggestion
```

Appreciate the custom hook, but we have `react-use` already so you can just use this 😄 

```react
import { useWindowSize } from "react-use";
const { width, height } = useWindowSize();
```

src/typescript/sdk/src/indexer-v2/types/index.ts
```diff
@@ -595,6 +595,18 @@ export const toUserPoolsRPCResponse = (data: DatabaseJsonType["user_pools"]) =>
...
+const unq64 = (n: number) => n / 2 ** 64;
+
+export const toPriceFeedRPCResponse = (data: DatabaseJsonType["price_feed"]) => ({
+  marketID: data.market_id,
+  symbolBytes: data.symbol_bytes,
+  symbolEmojis: data.symbol_emojis,
+  marketAddress: data.market_address,
+  openPrice: unq64(data.open_price_q64),
+  closePrice: unq64(data.close_price_q64),
+  deltaPercentage: (1 - unq64(data.close_price_q64) / unq64(data.open_price_q64)) * 100,
```
xbtmatt, line 607:
```suggestion
  deltaPercentage: Big(1)
    .minus(q64ToBig(data.close_price_q64).div(q64ToBig(data.open_price_q64)))
    .mul(100);
```

Since these are numeric types let's be safe and use `Big` for this

src/typescript/sdk/src/indexer-v2/types/index.ts
```diff
@@ -595,6 +595,18 @@ export const toUserPoolsRPCResponse = (data: DatabaseJsonType["user_pools"]) =>
   dailyVolume: BigInt(data.daily_volume),
 });
 
+const unq64 = (n: number) => n / 2 ** 64;
```
xbtmatt, line 598: And then remove this

src/typescript/sdk/src/indexer-v2/types/json-types.ts
```diff
@@ -357,6 +358,12 @@ export type DatabaseJsonType = {
       ProcessedFields &
       UserLPCoinBalance & { daily_volume: Uint128String }
   >;
+  [DatabaseRpc.PriceFeed]: Flatten<
+    Omit<MarketAndStateMetadata, "bump_time" | "market_nonce" | "trigger"> & {
+      open_price_q64: number;
+      close_price_q64: number;
```
xbtmatt, line 364: Maybe we can use something like what you did with `Decimal` and do`export type UnsizedNumericString = string;` for these? in `types.ts`

Would be nice so we can remember what they are

xbtmatt (changes requested):
Just the changes I mentioned- mostly all pretty straightforward

src/typescript/frontend/src/hooks/use-node-dimensions.ts
xbtmatt: Discussed on call, we'll fix this in another PR! Too difficult to get it working right now since it requires reactively updating the element based on the window size, not just getting the element width

xbtmatt (approved):
Awesome, looks great!

PR #257, opened 2024-09-23
[ECO-2103] Add inputs for min out amounts

CRBl69:
# Description

This PR adds an advanced mode for swaps and liquidity provisioning.

The advanced mode adds a new field titled "Minimum output amount".

The user can then set a minimum output amount.

Whenever an other input changes, the minimum output amount is automatically updated.

image

image

image

image

image

image

image

image

image

# Testing

See vecrel.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/components/info/index.tsx
+++ b/src/typescript/frontend/src/components/info/index.tsx
@@ -2,14 +2,14 @@ import React from "react";
 import Image from "next/image";
 
 import info from "../../../public/images/infoicon.svg";
-import Popup from "components/popup";
+import { PixelPopup } from "components/popup";
 
 const Info: React.FC<
-  React.PropsWithChildren<{ popupClassName?: string; imageClassName?: string }>
+  { popupClassName?: string; imageClassName?: string, children: string }
 > = ({ imageClassName, children, popupClassName }) => (
-  <Popup className={`max-w-[300px] ${popupClassName}`} content={children}>
+  <PixelPopup className={`max-w-[300px] ${popupClassName}`} content={children}>
     <Image src={info} alt="info" className={imageClassName} />
-  </Popup>
+  </PixelPopup>
 );
 
 export default Info;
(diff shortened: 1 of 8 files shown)


CRBl69:
I updated it to use percentages.

Now, whenever a user chooses to use percentages, this is what it looks like:

image

When clicking on the percent sign, it switches to this:

image

Note that the newly displayed value upon mode change is calculated as `prevPercentage * outputAmount` or `prevMinOut / outputAmount * 100`.

Also, percentage mode is enabled by default

src/typescript/frontend/src/lib/server-env.ts
```diff
@@ -42,8 +42,8 @@ export const REVALIDATION_TIME: number = Number(process.env.REVALIDATION_TIME);
...
-if (APTOS_NETWORK !== Network.LOCAL && EMOJICOIN_INDEXER_URL.includes("localhost")) {
-  throw new Error(
-    `APTOS_NETWORK is ${APTOS_NETWORK} but the indexer processor url is ${EMOJICOIN_INDEXER_URL}`
-  );
-}
+//if (APTOS_NETWORK !== Network.LOCAL && EMOJICOIN_INDEXER_URL.includes("localhost")) {
+//  throw new Error(
+//    `APTOS_NETWORK is ${APTOS_NETWORK} but the indexer processor url is ${EMOJICOIN_INDEXER_URL}`
+//  );
+//}
```
xbtmatt, line 49: Can undo this and rebase

xbtmatt (commented):
The design for the swap page is awesome! Personally I think it fits really well, maybe we can do it for the liquidity page, too?

I'm also open to the toggle on both as well, but I think it needs a label, not just a hover tooltip.

But let's go with the `+` design on both first. It's really intuitive.

Also I really like that you made a % input, was that in the figma or did you just add it? Looks great!

CRBl69:
There is no Figma for this. Just freestyled.

xbtmatt (commented):
Closing this PR in favor of waiting for Kirsten to make a new design for this.

Just want to note that I thought it was nice, and liked the initiative you took freestyling it :)

PR #262, opened 2024-09-25
[ECO-2204] Update "i" icon design on launch page

CRBl69:
# Description

On the pools page:

This PR moves the "i" icon to the left of the text and keeps the right alignment.

It also makes it slightly smaller.

# Testing

See vecrel.

# Checklist


xbtmatt (approved):
Note that I changed it back to having the `i` after since it looked much better. I used the new prop to make it slightly smaller though, removed the colon, and removed the uppercase- it looks great now!

PR #263, opened 2024-09-26
[ECO-2209] Make insufficient balance for market registration clearer

xbtmatt:
# Description

Change the launch emojicoin button to say:

- [x] `Invalid Input` if it's an invalid input of emojis somehow (It shouldn't be able to do this)
- [x] `Launch Emojicoin` otherwise
- [x] Add a ✅ or ❌ if the balance appears sufficient/insufficient, respectively
- [x] Fix the `invalid` logic to make sure all emojis are valid, not just checking for one being valid

image

image

# Testing

Manual testing.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/components/pages/launch-emojicoin/memoized-launch/components/launch-or-goto.tsx
+++ b/src/typescript/frontend/src/components/pages/launch-emojicoin/memoized-launch/components/launch-or-goto.tsx
@@ -10,13 +10,11 @@ export const LaunchButtonOrGoToMarketLink = ({
   onWalletButtonClick,
   registered,
   invalid,
-  insufficientBalance,
   geoblocked,
 }: {
   onWalletButtonClick: () => void;
   registered?: boolean;
   invalid: boolean;
-  insufficientBalance: boolean;
   geoblocked: boolean;
 }) => {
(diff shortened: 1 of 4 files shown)


src/typescript/frontend/src/components/pages/launch-emojicoin/memoized-launch/components/launch-or-goto.tsx
```diff
@@ -39,13 +41,19 @@ export const LaunchButtonOrGoToMarketLink = ({
...
             scrambleProps={scrambleProps}
           >
-            {t("Launch Emojicoin")}
+            {t(
+              insufficientBalance
+                ? "Insufficient balance"
+                : invalid
+                  ? "Invalid input"
+                  : "Launch Emojicoin"
+            )}
```
CRBl69, line 56: I think we settled on not blocking the user from registering a market whenever we think the balance is insufficient.

xbtmatt: Discussed on call, I think I'll do a pop-up if it's insufficient, rather than disabling it

CRBl69 (approved):

PR #268, opened 2024-09-27
[ECO-1955] Update loading animation with hearts

CRBl69:
# Description

This PR updates the loading animation to match the general theme of emojicoin.fun.

It uses a cool animation with colored hearts.

# Testing

See vercel preview.

Go to `/loading` to see the loading animation endlessly.

# Checklist


xbtmatt (commented):
As discussed on call, let's make sure the emojis are all centered for the one that the other emojis are circling.

You can probably visually test this easily by cycling through different random emojis *really* fast, and seeing which ones change the overall position/size of the wrapper div

Ideally we'll have custom css for poorly fitting emojis. If this is too complicated for all the different emoji picker sets (Apple, Windows, Linux, etc) we can just use Apple emojis

xbtmatt (approved):

PR #269, opened 2024-09-27
[ECO-2216] Add new button scale "xl" and set core interaction buttons to be "xl"

CRBl69:
# Description

Buttons were found to be too small.

This PR defines a new button size standard: "xl" (in addition to "sm" and "lg" which were already present).

The new standard is used in core interaction buttons, like swap, provide liquidity, or launch market.

# Testing

See vercel.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/components/button/index.tsx
+++ b/src/typescript/frontend/src/components/button/index.tsx
@@ -8,7 +8,7 @@ import SpinnerIcon from "components/svg/icons/Spinner";
 import { Text } from "components/text";
 import { FlexGap } from "@containers";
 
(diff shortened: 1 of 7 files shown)


xbtmatt (commented):
Where was this requested btw? Just curious, I personally think the button size is fine, but I suppose it could be small on mobile.

There are some inconsistencies with the sizes here- we'll have to look through everything meticulously and make sure they all make sense. Some of them won't fit

CRBl69:
There is a link to the issue in the linear task.

xbtmatt (commented):
Closing this in favor of the new button design, keeping them all the same size

PR #270, opened 2024-09-30
[ECO-2222] Clean up the localnet setup process

xbtmatt:
# Description

- [x] Add `bc` to the `localnet` service so `check-if-stale.sh` actually works
- [x] Remove `@aptos-labs/aptos-cli` from the dev dependencies since it's fickle and we don't use it anymore anyway
- [x] Fix the need to use `sudo` to remove the localnet data directory: we can do this by removing the `src/docker/localnet/.aptos` folder by starting a docker service, bind-mounting it one directory above, then removing `.aptos`
- [x] Add instructions for setting the `localnet` up in a Docker Desktop on Windows Subsystem for Linux (WSL) environment

# Testing

this PR.

# Checklist

### DIFF
--- a/src/docker/utils/prune.sh
+++ b/src/docker/utils/prune.sh
@@ -132,11 +132,12 @@ docker rm -f $api 2>/dev/null
 docker volume rm -f $postgres-data
 
 if [ -n "$reset_localnet" ]; then
-	if [ -n "$CI" ]; then
-		sudo rm -rf $docker_dir/localnet/.aptos || true
-	else
-		rm -rf $docker_dir/localnet/.aptos || true
-	fi
+	# In order to avoid having to run `sudo rm -rf src/docker/localnet/.aptos`,
+	# we can bind-mount a volume one directory above `localnet/.aptos`, aka
+	# `localnet`, effectively giving the ephemeral container ownership of
+	# everything in that directory. Then we delete the localnet data, aka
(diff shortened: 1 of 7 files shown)


src/docker/utils/prune.sh
```diff
@@ -132,11 +132,12 @@ docker rm -f $api 2>/dev/null
...
-	if [ -n "$CI" ]; then
-		sudo rm -rf $docker_dir/localnet/.aptos || true
-	else
-		rm -rf $docker_dir/localnet/.aptos || true
-	fi
+	# In order to avoid having to run `sudo rm -rf src/docker/localnet/.aptos`,
+	# we can bind-mount a volume one directory above `localnet/.aptos`, aka
+	# `localnet`, effectively giving the ephemeral container ownership of
+	# everything in that directory. Then we delete the localnet data, aka
+	# `.aptos`, from within the ephemeral container.
```
xbtmatt, line 139:
```suggestion
	# In order to avoid having to run `sudo rm -rf src/docker/localnet/.aptos`,
	# we can run an ephemeral `docker` container that bind-mounts `localnet`.
	# Bind-mounting the parent of `.aptos` gives the container the right to
	# delete it.
```

alnoki (approved):

CRBl69 (approved):

PR #282, opened 2024-10-04
[ECO-2192] Update liquidity page

CRBl69:
# Description

This PR implements the missing features of the pool page as well as adopt the new processor fixes related to this.

- [x] Narrowed the `normalizeHex` import in `verify.ts` so that we don't see a warning when running `pnpm run build`, since it was leaking `@aptos-labs/ts-sdk` dependencies into the Edge Runtime (since `verify.ts` is middleware)
- [x] Added `REVALIDATION_TIME` to `pools/api.ts`, otherwise it was quite literally caching query results for a *year*
- [x] Call the `user_pools` rpc function call as a `GET` request instead of `POST`, because it simplies the API gateway authorization process for requests (otherwise we'd have to add a new API endpoint in API gateway)
- [x] Default to `1` if `daily_tvl_per_lp_coin_growth` is empty, since that results in `0.00%` instead of `-100%`
- [x] If `daily_tvl_per_lp_coin_growth` is missing, use `daily_tvl_per_lp_coin_growth_q64`. We can roll this change back once the `main` deployment is updated with the new images

# Testing

Unit and e2e tests:

- [x] Added unit tests for parsing to ensure that any incorrect behavior in parsing is caught there and not in the actual fetch requests
- [x] Added an e2e test for checking that the schema has been reloaded post-migration


.github/workflows/push-broker.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
         context: 'src/rust'
         file: 'src/rust/broker/Dockerfile'
         labels: '${{ steps.metadata.outputs.labels }}'
-        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
+        platforms: 'linux/amd64'
```
xbtmatt, line 28:
```suggestion
        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS}}'
```

.github/workflows/push-broker.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
         context: 'src/rust'
         file: 'src/rust/broker/Dockerfile'
         labels: '${{ steps.metadata.outputs.labels }}'
-        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
+        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS}}'
```
xbtmatt, line 28:
```suggestion
        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
```

src/typescript/sdk/tests/e2e/broker/websockets.test.ts
```diff
@@ -43,7 +43,7 @@ const customWaitFor = async (condition: () => boolean) =>
   waitFor({
     condition,
     interval: 10,
-    maxWaitTime: 5000,
+    maxWaitTime: 10000,
```
xbtmatt, line 46:
```suggestion
    maxWaitTime: 5000,
```

xbtmatt (approved):
Just want to note the changes I made here:

- [ ] Added unit tests for parsing to ensure that any incorrect behavior in parsing is caught there and not in the actual fetch requests
- [ ] Added an e2e test for checking that the schema has been reloaded post-migration
- [ ] Narrowed the `normalizeHex` import in `verify.ts` so that we don't see a warning when running `pnpm run build`, since it was leaking `@aptos-labs/ts-sdk` dependencies into the Edge Runtime (since `verify.ts` is middleware)
- [ ] Added `REVALIDATION_TIME` to `pools/api.ts`, otherwise it was quite literally caching query results for a *year*
- [ ] Call the `user_pools` rpc function call as a `GET` request instead of `POST`, because it simplies the API gateway authorization process for requests (otherwise we'd have to add a new API endpoint in API gateway)
- [ ] Default to `1` if `daily_tvl_per_lp_coin_growth` is empty, since that results in `0.00%` instead of `-100%`
- [ ] If `daily_tvl_per_lp_coin_growth` is missing, use `daily_tvl_per_lp_coin_growth_q64`. We can roll this change back once the `main` deployment is updated with the new images

PR #283, opened 2024-10-04
[ECO-2243] Fix NextJS config and cache

CRBl69:
# Description

This PR fixes the NextJS caching configuration and routing configuration:

- Caching:
    - `revalidate` is hard coded to `2` as NextJS does not support setting `revalidate` to dynamic values.
    - `force-static` and `force-dynamic` are removed since they're mostly useless.
- Routing:
    - `app/page.tsx` and `app/home/page.tsx` were duplicates. With the new configuration changes, `app/page.tsx` is removed (actually, replaced with a loading page just in case anyone manages to land here) and `/` redirects to `/home`.

# Testing

See vercel (also vercel logs for cache).

# Checklist


xbtmatt (commented):
I'm a little confused why we are switching from forcing all requests to be dynamic to by default caching everything. As far as I understand, caching requests will cache them indefinitely unless revalidate is set. Setting it to 1 should work for fetches on that page, but I'm not sure about requests from any other part of the route tree or layout.

Also, did you test the variable revalidate amount and find that it didn't matter what you set it as? Because I have seen the value changed in the `.next/fetch/*` files. And the docs here suggest it's valid to use a number: https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#revalidate

I believe that it's only ignored in the `edge` runtime, but I believe that's only used in our middleware right now (I could be wrong about this).

In general though there are many cases where we really don't want to cache at all, and some where we want to cache much longer than 1 second. I think if we want to implement caching in any capacity, we should have e2e tests for it and a thorough investigation into how it works.

That is, until we have an extensive understanding of how caching works, we should probably just default to no caching at all, and it sems like this is the opposite of that. 

In other words, I think we should default to `const fetchCache: 'default-no-store'` if we're going to do anything

src/typescript/frontend/next.config.mjs
alnoki, line 1: - [ ] Also verify `"use server"`effects
- [ ] Mark this resolved

xbtmatt (commented):
Just commenting here to reiterate that the main thing to test is all server fetches used in files with the `"use server";` directive at the top, since they are all essentially overloaded nextjs fetch functions.

The main thing is to see what router path they're included in, since the file can be completely disparate from the file-based directory that normally resolves the path for caching

src/typescript/frontend/src/app/home/loading.tsx
alnoki, line 1: - [ ] Note typical guidance `to not use server actions for anything other than mutations` may be invalid
- [ ] See https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations
- [ ] Note that candlestick fetches are done through the charting library
- [ ] Incorporate into `route.ts` like on pools page
- [ ] Resolve this comment once all items addressed

src/typescript/frontend/src/components/charts/PrivateChart.tsx
```diff
@@ -145,119 +145,117 @@ export const Chart = async (props: ChartContainerProps) => {
...
+              high: 0,
+              low: 0,
+              close: 0,
+              volume: 0,
+            });
+          } else {
+            onHistoryCallback([], {
+              noData: true,
+            });
+            return;
```
xbtmatt, line 252:
```suggestion
```

I think this is pointless actually, right? Ignore me if not, but feel free to change + resolve this comment either way

src/typescript/frontend/src/components/charts/PrivateChart.tsx
```diff
@@ -145,119 +145,117 @@ export const Chart = async (props: ChartContainerProps) => {
...
+        const periodDurationSeconds = (periodDuration / PeriodDuration.PERIOD_1M) * 60;
+        const end = new Date(to * 1000);
+        // The start timestamp is rounded so that all the people who load the webpage at a similar time get served
+        // the same cached response.
+        const start = from - (from % periodDurationSeconds);
+        const data: PeriodicStateEventModel[] = await fetch(
+          `/candlesticks?marketID=${props.marketID}&start=${start}&period=${period}&limit=500`
+        )
+          .then((res) => res.text())
+          .then((res) => parseJSON(res));
```
xbtmatt, line 167: Is there a reason you got rid of the try/catch here?

src/typescript/frontend/src/components/charts/PrivateChart.tsx
```diff
@@ -145,119 +145,117 @@ export const Chart = async (props: ChartContainerProps) => {
...
+        const period = ResolutionStringToPeriod[resolution.toString()];
+        const periodDuration = periodEnumToRawDuration(period);
+        const periodDurationSeconds = (periodDuration / PeriodDuration.PERIOD_1M) * 60;
+        const end = new Date(to * 1000);
+        // The start timestamp is rounded so that all the people who load the webpage at a similar time get served
+        // the same cached response.
+        const start = from - (from % periodDurationSeconds);
+        const data: PeriodicStateEventModel[] = await fetch(
+          `/candlesticks?marketID=${props.marketID}&start=${start}&period=${period}&limit=500`
+        )
```
xbtmatt, line 165: I think you should use the `searchParams` fetch field here, it's difficult to read/parse a raw URL string. There should be one where you can specify the query params as `key: value`s

src/typescript/frontend/src/components/loading.tsx
```diff
@@ -1,3 +1,5 @@
+"use client";
+
```
xbtmatt, line 2: How come you added this here? I think it's a client component by necessity and thus by default. Just curious

src/typescript/frontend/src/app/page.tsx
```diff
@@ -1,45 +1,6 @@
...
-      page={page}
-      sortBy={sortBy}
-      searchBytes={q}
-      geoblocked={geoblocked}
-      priceFeed={priceFeed}
-    />
-  );
+export default function Home() {
+  return <LoadingComponent />;
 }
```
xbtmatt, line 6: What exactly is going on here? Did you change this to a loading page because it's just a redirect now?

xbtmatt (commented):
Just the questions I had in the PR, WRT the caching behavior it's really difficult to tell if it's even working correctly or not. I'm a little uneasy about actually changing the caching behavior in general- but I know we've discussed this before.

It's been a bit since we talked about it so could you remind me again why exactly we're changing the caching behavior? I can't exactly remember why we decided to change it in the first place, since it seemed to have been working

src/typescript/frontend/src/components/loading.tsx
CRBl69: It was such a while ago I can't exactly remember. Probably I got an error without it.

src/typescript/frontend/src/app/page.tsx
CRBl69: I think we talked about this. I'm redirecting all requests from `/` to `/home` so that we don't have a duplicated home page. This actually introduced an error where the `/home` was working properly but `/` not. I left a loading component here just in case this page somehow renders, but it should never happen.

src/typescript/frontend/src/components/charts/PrivateChart.tsx
CRBl69: Not sure what this is supposed to do TBH.

src/typescript/frontend/next.config.mjs
alnoki, line 1: Per 2024-10-31 eng call

- [ ] Use the site and verify that the data displayed is correct
- [ ] Submit trade, go away, then go back to page, and data should be there
- [ ] Resolve this comment once verified

CRBl69:
TODO:

- [ ] Fix unstable cache not being able to serialize bigints

xbtmatt (commented):
Noting here that to review this fully I will:

- [ ] Check vercel cache when testing this
- [ ] Make sure this doesn't break/make the UX/UI much worse

src/typescript/frontend/playwright.config.js
```diff
@@ -24,7 +24,7 @@ export default defineConfig({
     baseURL: "http://127.0.0.1:3001",
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: "retain-on-first-failure",
+    trace: "on-first-retry",
```
xbtmatt, line 27:
```suggestion
    trace: "retain-on-first-failure"",
```

src/typescript/frontend/playwright.config.js
```diff
@@ -24,7 +24,7 @@ export default defineConfig({
     baseURL: "http://127.0.0.1:3001",
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: "retain-on-first-failure",
+    trace: "retain-on-first-failure"",
```
xbtmatt, line 27:
```suggestion
    trace: "retain-on-first-failure",
```

xbtmatt (commented):
Edit: Moved the original comment here in favor of saying it while requesting changes

xbtmatt (changes requested):
I posted a lengthy review of nextjs caching and this PR in general on Slack, but the jist of it is that I don't think we should implement most of the changes introduced in this PR because they're misaligned with nextjs' caching mechanisms, although I definitely agree with the underlying motivation (we need caching)

I explained the follow up solution/changes we need on slack

xbtmatt (approved):
This is working pretty well! Thanks for implementing this- I appreciate you getting this stuff to work with a more simple/elegant solution than using `unstable_cache` everywhere

PR #285, opened 2024-10-07
[ECO-2233] Add price feed e2e test

CRBl69:
# Description

This PR adds a price feed e2e test.

It also adds a new env var (`DB_URL`) to allow DB inserting in E2E tests.


xbtmatt (changes requested):
Let's add a test case for the new logic in the indexer processor- the new market swap where there's no trade more than 24 hours ago but there is another one within the past 24 hours

Appreciate the insertion stuff => `.sql` file clean-up!

alnoki:
Marking as draft to prevent merge into base branch, until base branch becomes `main`

xbtmatt (approved):
I don't see the actual price feed display function (the percentage calculation) called in any tests here, but it seems you are just verifying that the open/close prices are as expected.

If it's easy to add in, could you just add another test that verifies that the price feed percentage value returned to the actual frontend is being calculated correctly?

Should just be a few extra lines of code, you don't even have to make it an e2e test, since you've already verified the e2e part of it

Besides that, looks good!

src/typescript/sdk/tests/e2e/queries/price-feed/price-feed.test.ts
alnoki, line 1: - [ ] Incorporate new test described at https://github.com/econia-labs/emojicoin-dot-fun/pull/285#pullrequestreview-2406829751
- [ ] Mark this as resolved

PR #293, opened 2024-10-09
[ECO-2244] Add js/ts linting to pre-commit

CRBl69:
# Description

This PR adds js/ts linting to pre-commit.

### DIFF
--- a/cfg/pre-commit-config.yaml
+++ b/cfg/pre-commit-config.yaml
@@ -109,6 +109,16 @@ repos:
     name: 'mypy'
     types:
     - 'python'
+  -
+    entry: |
+      pnpm run lint
+    files: 'src/typescript'
+    id: 'pnpm'
+    language: 'node'
+    name: 'lint typescript'
+    pass_filenames: false
+    types:
+    - 'directory'
   repo: 'local'
 -
   hooks:
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -7,6 +7,7 @@
 env:
   POETRY_VERSION: '1.8.2'
   PYTHON_VERSION: '3.10'
+  TS_DIR: 'src/typescript'
 jobs:
   pre-commit:
     runs-on: 'ubuntu-latest'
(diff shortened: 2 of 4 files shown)


cfg/pre-commit-config.yaml
alnoki, line 1: - [ ] Verify that this hook encapsulates everything done in the existing TS CI
- [ ] Delete both actions so that they no longer run in duplicate on GitHub actions
- [ ] Mark this comment resolved

alnoki (approved):

cfg/pre-commit-config.yaml
```diff
@@ -109,6 +109,16 @@ repos:
...
+  -
+    entry: |
+      pnpm run lint
+    files: 'src/typescript'
+    id: 'pnpm'
+    language: 'node'
+    name: 'lint typescript'
+    pass_filenames: false
+    types:
+    - 'directory'
```
xbtmatt, line 121: does this run on all files? I can't find `language: node` in `pre-commit` config on the site

xbtmatt (changes requested):
The request for changes isn't really due to a change you made, but I'd like to get it in in this PR, so noting it here:

After using `prettier` in `eslint` for a bit now I think we should keep them separate. The constant red squiggly lines in vscode are really bothering me so I've begun turning the rule off for prettier in eslint, which means it won't auto-run it anymore when running lint:fix.

This also means we need to run `prettier --check` in pre-commit, which is fine, but you'll have to add it to this PR:

1. Remove the `    "prettier/prettier": ["error"],` rule from `.eslintrc.js` in `sdk` and `frontend`
2. Run `pnpm run lint:fix` and `pnpm run format`, and then `pnpm run format` and `pnpm run lint:fix`  and note if there are no changes when the order of running them changes. If so, we just need to make sure pre-commit runs them in a specific order each time. I think ideally I'd prefer `lint:fix` and *then* `prettier` but I think they should be equivalent.
3. Use `pnpm run format:check` to run prettier in `pre-commit`. It fails if there are changes necessary without making changes

cfg/pre-commit-config.yaml
CRBl69: This only run once. Since `pnpm run lint` already checks all the files.

cfg/pre-commit-config.yaml
CRBl69: I'll note that we do not want to run e2e in pre-commit, only linting and formatting. I'll let e2e in CI.

xbtmatt (approved):
Looks good now!

PR #299, opened 2024-10-13
[ECO-2246] Add QoL `EmojicoinClient` class to make testing and generating data easier

xbtmatt:
# Description

There's tons of repeated code that we can get rid of by making a simple client class to submit transactions that derives values and stores things like the `Aptos` client in it. This will make using the SDK significantly simpler and pave the way for more QoL improvements for generating data and writing tests.

- [x] Add the client
- [x] Add it to new tests to avoid regressions

# Testing

New tests.

# Checklist


src/typescript/sdk/src/client/emojicoin-client.ts
```diff
@@ -0,0 +1,332 @@
...
+      },
+    };
+  }
+
+  async buy(
+    swapper: Account,
+    emojis: SymbolEmoji[],
+    args: {
+      inputAmount: Uint64;
+      minOutputAmount: Uint64;
```
CRBl69, line 152: How about making this optional with a default value of either 99% or just 1. Since we want to use it for easy testing, it would be easier to pass fewer args, and we will rarely need to pass a minimum output amount.

src/typescript/sdk/src/client/emojicoin-client.ts
```diff
@@ -0,0 +1,332 @@
...
+      integrator: AccountAddressInput;
```
CRBl69, line 153: Same here, how about abstracting this away ? Maybe get the default one from env if none is passed ?

src/typescript/sdk/src/client/emojicoin-client.ts
```diff
@@ -0,0 +1,332 @@
...
+      integratorFeeRateBPs: number;
```
CRBl69, line 154: Same here.

src/typescript/sdk/src/client/emojicoin-client.ts
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  async sell(
+    swapper: Account,
+    emojis: SymbolEmoji[],
+    args: {
+      inputAmount: Uint64;
+      minOutputAmount: Uint64;
+      integrator: AccountAddressInput;
+      integratorFeeRateBPs: number;
```
CRBl69, line 168: Same as for the above function, and same for all similar places.

src/typescript/sdk/src/client/emojicoin-client.ts
```diff
@@ -0,0 +1,332 @@
...
+    },
+    options?: Options
+  ) {
+    return await this.swap(swapper, emojis, { ...args, isSell: true }, options);
+  }
+
+  private async swap(
+    swapper: Account,
+    emojis: SymbolEmoji[],
+    args: ExtraSwapArgs,
```
CRBl69, line 178: For consistency, either put all `args` without a type (i.e. replace this with `args: {...}`), or put all `args` with a dedicated type (i.e. replace `args: {...}` in other functions with `args: Type`).

CRBl69 (changes requested):

src/typescript/sdk/src/client/emojicoin-client.ts
xbtmatt: So actually my first instinct was to do that- I initially made the class with a bunch of default args and assumptions because it makes it even easier to use (like you're saying, fewer args, easier to read, etc).

But I think it's better if we keep this specifically as a class to simplify transaction submission without making any assumptions and make another class extended from this one specifically as a test helper. Partly to separate concerns (general helper class vs test-only helper class) and partly to make it easier to test.

I can add the test-only class today and that should cover streamlining some of the unnecessary abstractions in our tests

src/typescript/sdk/src/client/emojicoin-client.ts
xbtmatt: I do agree here that the base class (vs the test-only class, as mentioned in the other comment) could pass in an `integrator` and `integratorFeeRateBPs` upon instantiation to avoid having to specify them later.

Possibly some other fields that would mean having to specify fewer args as well

CRBl69 (approved):

PR #301, opened 2024-10-15
[ECO-2280] Replace 00 on home page

CRBl69:
# Description

This PR aims to replace the confusing `00` on the home page.

Here are some options:

## Option A0 - 🔥

image

Personal opinion: this one feels the most confusing to me. It seems like it's the market name to me.

## Option A1 - HOT 🔥

image

Personal opition: not a fan of the emoji, but this is the best compromise between having and emoji and clarity.

## Option A2 - HOT RIGHT NOW 🔥

image

Personal opinion: maybe a bit too much.

## Option A3 - FEATURED 🔥

image

Personal opition: just like option A1, but "featured" might give the false idea that the market is somehow officially endorsed.

## Option B1 - HOT

image

Personal opinion: I like the look, very clean, but not very visible.

## Option B2 - HOT RIGHT NOW

image

Personal opinion: sometimes, you don't need a reason to hate something.

## Option B3 - FEATURED

image

Personal opinion: Looks the cleanest out of them all, but has the endorsement issue.

### DIFF
--- a/src/typescript/frontend/src/components/pages/home/components/main-card/MainCard.tsx
+++ b/src/typescript/frontend/src/components/pages/home/components/main-card/MainCard.tsx
@@ -13,6 +13,7 @@ import planetHome from "../../../../../../public/images/planet-home.png";
 import { emojiNamesToPath } from "utils/pathname-helpers";
 import { type HomePageProps } from "app/home/HomePage";
 import "./module.css";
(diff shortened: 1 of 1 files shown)


xorkiki:
I vote "Option A1"

xbtmatt (approved):
Let's go with the Hot 🔥, just add a space between the Hot and 🔥

`Hot 🔥`

PR #304, opened 2024-10-17
[ECO-2195] Add market discovery rain

CRBl69:
# Description

This PR makes it rain.

The launch page now rains with random unregistered market names.

Clicking on one will pre-fill the input.

# Testing

See vecrel.

### DIFF
--- a/src/typescript/frontend/src/app/launch/page.tsx
+++ b/src/typescript/frontend/src/app/launch/page.tsx
@@ -2,11 +2,13 @@ import { REVALIDATION_TIME } from "lib/server-env";
 import ClientLaunchEmojicoinPage from "../../components/pages/launch-emojicoin/ClientLaunchEmojicoinPage";
 import { isUserGeoblocked } from "utils/geolocation";
 import { headers } from "next/headers";
+import { fetchRandomNames } from "@/queries/launch";
 
 export const revalidate = REVALIDATION_TIME;
 export const dynamic = "force-static";
 
 export default async function LaunchEmojicoinPage() {
+  const randomNames = await fetchRandomNames({});
   const geoblocked = await isUserGeoblocked(headers().get("x-real-ip"));
-  return <ClientLaunchEmojicoinPage geoblocked={geoblocked} />;
+  return <ClientLaunchEmojicoinPage geoblocked={geoblocked} randomNames={randomNames} />;
 }
--- a/src/typescript/frontend/src/components/pages/launch-emojicoin/EmojiRain.tsx
+++ b/src/typescript/frontend/src/components/pages/launch-emojicoin/EmojiRain.tsx
@@ -1,0 +1,69 @@
+import { type symbolBytesToEmojis } from "@sdk/emoji_data";
+import { animate, motion, stagger } from "framer-motion";
+import { useEffect, useMemo } from "react";
+import { useWindowSize } from "react-use";
+
+export function EmojiRainDrop({
+  name,
+  onClick,
+}: {
+  name: ReturnType<typeof symbolBytesToEmojis>;
+  onClick?: (data: ReturnType<typeof symbolBytesToEmojis>) => void;
+}) {
+  const { width } = useWindowSize();
+  const initialX = useMemo(() => -10 - 100, []);
+  const y = useMemo(() => Math.random() * (width - 20) + 10, [width]);
+
+  return (
+    <motion.div
+      onClick={() => onClick && onClick(name)}
+      className="absolute emoji-rain-drop z-10 flex flex-col select-none cursor-pointer"
+      style={{
+        left: y,
+        fontSize: "2em",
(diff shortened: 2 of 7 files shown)


src/typescript/frontend/src/app/launch/page.tsx
alnoki, line 1: - [ ] Check all Linear boxes

src/typescript/sdk/src/indexer-v2/types/json-types.ts
```diff
@@ -364,6 +365,7 @@ export type DatabaseJsonType = {
       close_price_q64: Uint64String;
     }
   >;
+  [DatabaseRpc.RandomNames]: { emojis: string };
```
xbtmatt, line 368: Is there a more specific format to expect? It looks like it might be in the form of hex with an `x` prepended? It'd be nice to have it here just so it's clear what to expect when calling this function

You can do something like `x${string}`

xbtmatt (commented):
Visiting the launch page doesn't seem to show anything for me- is it working? I am just going to the launch page. I waited for a few minutes and still don't see anything

CRBl69:
@xbtmatt nothing is showing up because the backend is not up to date and does not have the `random_symbols` function present.

src/typescript/sdk/src/indexer-v2/queries/app/launch.ts
```diff
@@ -0,0 +1,14 @@
+import "server-only";
```
xbtmatt, line 1:
```suggestion
if (process.env.NODE_ENV !== "test") {
  require("server-only");
}
```

src/typescript/sdk/src/indexer-v2/types/index.ts
```diff
@@ -596,6 +601,9 @@ export const toPriceFeed = (data: DatabaseJsonType["price_feed"]) => {
   };
 };
 
+export const toRandomSymbolsRPCResponse = (data: DatabaseJsonType["random_symbols"]) =>
+  symbolBytesToEmojis("0" + data.emojis.substring(1));
```
xbtmatt, line 605: What exactly is this doing? Is this replacing the `\x` from the beginning of the string with `0x`? Let's put that in a comment, otherwise this is a bit confusing

src/typescript/sdk/src/indexer-v2/types/json-types.ts
```diff
@@ -288,6 +288,7 @@ export enum TableName {
 export enum DatabaseRpc {
   UserPools = "user_pools",
   PriceFeed = "price_feed",
```
xbtmatt, line 290:
```suggestion
```

src/typescript/sdk/src/indexer-v2/types/json-types.ts
```diff
@@ -365,6 +366,7 @@ export type DatabaseJsonType = {
       ProcessedFields &
       UserLPCoinBalance & { daily_volume: Uint128String }
   >;
+  [DatabaseRpc.RandomSymbols]: { emojis: string };
```
xbtmatt, line 369:
```suggestion
  [DatabaseRpc.RandomSymbols]: { emojis: `\x${string}` };
```

Let's update this if this is the case- not sure if it actually is though.

src/typescript/frontend/src/components/pages/launch-emojicoin/EmojiRain.tsx
```diff
@@ -0,0 +1,100 @@
...
+  index: number;
+  emojis: number;
+}) {
+  const { height, width } = useWindowSize();
+
+  // Position all emojis above the visible space.
+  const initialX = -200;
+
+  // Calculate a random position on the Y axis.
+  const y = useMemo(() => Math.random() * (width - 20) + 10, [width]);
```
xbtmatt, line 26: I think you have the variable names backwards? this is technically the initial `Y` position.

Also I think you should do this entirely with percentages and remove the window size hook- you can use `${percentage}vw` and `${percentage}vh` for the style

xbtmatt (changes requested):
Just the changes I mentioned- ideally we don't use `useWindowSize` because it's a fairly heavy hook afaik, but mostly you can do everything with pure css with `vw` and `vh`

Also, this is a fairly obtrusive design, I think we should add a switch toggle to enable it and have it disabled by default or it might end up just confusing/irritating people, unfortunately...

xbtmatt (commented):
Going to add the config for the switch to have this disabled by default but let users add it in when viewing the page

xbtmatt (commented):
Things to note discussed on call with Kirsten:

1. Make the "Select emoji" section more opaque than the rest, and have a higher z-index than the emojis fallling
2. Rotate the emojis like the other interaction?

More, but just wanted to write down really quick

alnoki:
Bumping stale label

PR #308, opened 2024-10-21
[ECO-2313] Fix input fields on market and pools pages

CRBl69:
# Description

This PR fixes input fields on the market and the pools pages.

Before, inputing small values would glitch the inputs.

Now, you can input values as small as 0.00000001, which is the smallest number you can trade.

# Testing

Input small values on the market and the pools pages in vercel.

### DIFF
--- a/src/typescript/frontend/src/components/pages/emojicoin/components/trade-emojicoin/SwapComponent.tsx
+++ b/src/typescript/frontend/src/components/pages/emojicoin/components/trade-emojicoin/SwapComponent.tsx
@@ -79,8 +79,8 @@ const inputAndOutputStyles = `
   border-transparent !p-0 text-white
 `;
 
-const APT_DISPLAY_DECIMALS = 4;
-const EMOJICOIN_DISPLAY_DECIMALS = 1;
+const INPUT_DISPLAY_DECIMALS = 8;
+const OUTPUT_DISPLAY_DECIMALS = 4;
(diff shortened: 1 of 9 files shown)


src/typescript/frontend/src/components/pages/emojicoin/components/trade-emojicoin/SwapComponent.tsx
alnoki, line 1: - [ ] Craft a modular, reusable component that can be used for text input cases, in a utility section of the repo or simlar

xbtmatt (changes requested):
So as we discussed on call I think this is fairly complex because we're doing manual form validation with a custom decimal display (from the Aptos coin module)

I think there's two ways we can go with this:

1. Add several unit tests specifically for the functions used in these components, and maybe a couple in playwright as well with actual input changes. This would be faster but it will continue to be difficult to maintain and update, although this likely won't change much again, + with tests it's easier to update safely.
2. Refactor this to use some form validation library and minimize the amount of code we use and make it modular for any future frontend stuff we do with input + display coin decimal formatting

If we go with #1 I think it's important to also document what all the various regex patterns do- at least the ones that take more than 5 seconds (for a human) to parse and understand

I'm comfortable with either one, so it's up to you to choose. if I had to choose, I'd say number 2

xbtmatt (approved):

PR #314, opened 2024-10-24
[ECO-2317] Refactor and rename tests

CRBl69:
# Description

This PR refactor, moves and renames some tests to make them and their scope more clear.

# Testing

See tests below, and see if their naming is better.

# Checklist


.github/workflows/frontend-tests.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 jobs:
-  test:
+  frontend-tests:
```
alnoki, line 3:
```suggestion
  playwright-tests:
```

alnoki (approved):

PR #323, opened 2024-11-04
[ECO-2363] Fix page number wrong on search

CRBl69:
# Description

On search, the number of pages was not correct. Now it is.

# Testing

Search for something, check page number.


xbtmatt (commented):
Reminder to fix the issue where the first number (e.g. 3 in 3/10) in the page numbers is wrong.

For example, if you're on the 3rd page and then search an emoji, it will still be on the 3rd page

xbtmatt (approved):

PR #327, opened 2024-11-05
[ECO-2370] Fix small bugs

CRBl69:
# Description

This PR fixes various small bugs inside atomic commits. See commits name for more details.

### DIFF
--- a/src/typescript/frontend/src/components/pages/home/components/emoji-table/index.tsx
+++ b/src/typescript/frontend/src/components/pages/home/components/emoji-table/index.tsx
@@ -23,7 +23,7 @@ import { encodeEmojis } from "@sdk/emoji_data";
 import { useEventStore, useUserSettings } from "context/event-store-context";
 import { LiveClientGrid } from "./AnimatedClientGrid";
 import useEvent from "@hooks/use-event";
-import { constructURLForHomePage, isHomePageURLDifferent } from "lib/queries/sorting/query-params";
+import { constructURLForHomePage } from "lib/queries/sorting/query-params";
 import { AnimatePresence, motion } from "framer-motion";
 import { EMOJI_GRID_ITEM_WIDTH } from "../const";
 import { useGridRowLength } from "./hooks/use-grid-items-per-line";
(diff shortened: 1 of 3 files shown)


src/typescript/frontend/src/components/pages/home/components/emoji-table/index.tsx
```diff
@@ -66,18 +66,13 @@ const EmojiTable = (props: EmojiTableProps) => {
...
       page: args?.page ?? page,
       sort: args?.sort ?? sort,
       searchBytes: encodeEmojis(args?.emojis ?? emojis),
     });
 
-    // Always push the new URL to the history, but only refresh if the URL has actually changed in a meaningful way.
     router.push(newURL.toString(), { scroll: false });
-    if (isHomePageURLDifferent(curr, newURL.searchParams)) {
-      router.refresh();
-    }
```
xbtmatt, line 80: This isn't doing nothing- `router.refresh()` triggers the server components to refetch their data. If you only do `router.push`, it will only change the URL in the browser's URL bar, but not actually reload the page.

The `isHomePageURLDifferent` is there to avoid a refresh when the home page is functionally the same (e.g. `/home` and `/` and `/home?sort_by=market_cap` are all the same page)

src/typescript/frontend/src/components/pages/home/components/table-card/TableCard.tsx
```diff
@@ -129,7 +127,14 @@ const TableCard = ({
...
   // Always show the left border when there's something in the search bar.
-  const borderLeftWidth = useMotionValue(curr.col === 0 ? 1 : anySearchBytes ? 1 : 0);
+  const borderLeftWidth = useMotionValue(curr.col === 0 ? 1 : 0);
+
+  useEffect(() => {
+    if (curr.col === 0) {
+      borderLeftWidth.set(1);
+    }
+    /* eslint-disable react-hooks/exhaustive-deps */
+  }, [curr]);
```
xbtmatt, line 137: I have a sneaking suspicion that this was only necessary because you changed the `router.refresh()` in the other file, can you make sure that that's not the case?

xbtmatt (changes requested):
I don't think we should make either of these changes- but if you have a reason to add them that addresses what I mentioned in the inline comments then let me know, otherwise I think we shouldn't change anything here and just close the PR

src/typescript/frontend/src/components/pages/home/components/table-card/TableCard.tsx
CRBl69: image

This is a screenshot from main, where that change is not present. See https://emojicoin-dot-fun-git-main-econia-labs.vercel.app/?q=0xf09f9095

src/typescript/frontend/src/components/pages/home/components/emoji-table/index.tsx
xbtmatt: Yes I checked to make sure it works still- I think I possibly added this while confused about how the router cache worked since we'd recently updated it to use a time of `0` instead of `30`, plus it may have also been bugged back then, too.

I checked if it sees new markets without the `router.refresh()` just by using push and it does, so you're right, all's good!

I agree about unused code- in which case I'll also remove `isHomePageURLDifferent` since it's only used in one place

src/typescript/frontend/src/components/pages/home/components/table-card/TableCard.tsx
xbtmatt: Awesome thanks for checking this- I thought I fixed this a while ago but I guess not, thanks for checking

Also what is that blue checkmark on the picture?

xbtmatt (approved):
Looks like both changes were good- apologies for jumping the gun on not getting this in!

PR #328, opened 2024-11-05
[ECO-2256] Fix line break on market with three emojis

CRBl69:
# Description

This PR fixes a bug where a line would break on the home page if a market had three emojis in its symbol.

This PR also adds the possibility to use `className` with the `Text` component and makes the `Text` component code more clear.

# Testing

See vercel (currently there is such a market on page 3 but this might change when you review).

# Checklist

### DIFF
--- a/src/typescript/frontend/src/app/global.css
+++ b/src/typescript/frontend/src/app/global.css
@@ -27,36 +27,6 @@
   --error: #f3263e;
 }
 
-.med-pixel-text {
-  font-size: 20px !important;
-  line-height: 25px;
-  font-family: var(--font-pixelar) !important;
-}
-
-@media screen and (min-width: 768px) {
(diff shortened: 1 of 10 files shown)


xbtmatt (commented):
Is there a reason you added the font family forma to the scales? That affects a lot of different elements by default, so it may change things unexpectedly if you didn't check every element on the site that uses the `scales` values

xbtmatt (approved):
Ah okay, I understand better what you did here, this does make more sense. Thanks for fixing this!

I added a conditional check for 3 emojis, making the font smaller in that case so that they all fit in the table card. I also moved some unnecessary global css stuff out of global.css and into a `module.css` file (for `.med-pixel-text` since it was only being used by the dropdown menu in one place)

PR #335, opened 2024-11-08
[ECO-2389] Render `/launch` as a static page

xbtmatt:
# Description

We could make `/launch` a completely static page, making it significantly faster to load, by removing the geoblock check and making it client-side while assuming the user is geoblocked while the geoblocked function check promise resolves.

- [ ] `const dynamic = "force-static"`  at the top of the file
- [ ] Check if the user is geoblocked with a react-query server action hook

# Testing

Check the build logs and ensure it's marked as a statically generated page in Vercel

# Checklist


src/typescript/frontend/src/hooks/use-is-user-geoblocked.ts
```diff
@@ -0,0 +1,15 @@
+import { useQuery } from "@tanstack/react-query";
+import { isUserGeoblockedServerAction } from "utils/server/geoblocked";
+
+const useIsUserGeoblocked = () => {
+  const { data } = useQuery({
+    queryKey: ["geoblocked"],
+    queryFn: () => isUserGeoblockedServerAction(),
+    staleTime: Infinity,
```
CRBl69, line 8: I think this should be something between 2 and 7 days. Ideally, when data is stale, it would return the stale data while new data is getting loaded, to avoid flashing the geo-blocked banner.

CRBl69 (changes requested):
We should only use one geo-blocking logic. If you think the hook works great, can you replace it everywhere ?

src/typescript/frontend/src/hooks/use-is-user-geoblocked.ts
xbtmatt: Addressed in 20575d7e

xbtmatt: Updated this to not show the banner unless they're explicitly geoblocked by adding an optional param to the hook that's false by default. That is, without specifying that you want to know if they're explicitly geoblocked (no nullish coalescing), we pass `{ explicitlyGeoblocked = true }` to the hook.

Otherwise, the data returned is `data ?? true` (assume they're geoblocked by default).

This lets us configure the banner to only show up if the user is explicitly geoblocked, but still block the user by default in other components without having to refactor those hook invocations.

cd7aab12cb0be9e1898ebae42b701e9e4fb8b2b4

src/typescript/frontend/src/components/header/wallet-button/ConnectWalletButton.tsx
```diff
@@ -25,12 +26,19 @@ export const ButtonWithConnectWalletFallback: React.FC<ConnectWalletProps> = ({
...
   const { connected, account } = useWallet();
   const { openWalletModal } = useWalletModal();
   const { t } = translationFunction();
+  const _geoblocked = useIsUserGeoblocked();
+  const geoblocked = useMemo(() => {
+    // For letting the user connect on the `/verify_status` page when `forceAllowConnect` is `true`,
+    // by only returning `geoblocked = true` if we're not force allow connecting and they're
+    // geoblocked.
+    return !forceAllowConnect && _geoblocked;
+  }, [forceAllowConnect, _geoblocked]);
```
CRBl69, line 41:
```suggestion
  const shouldBeGeoblocked = useIsUserGeoblocked();
  const geoblocked = useMemo(() => {
    // For letting the user connect on the `/verify_status` page when `forceAllowConnect` is `true`,
    // by only returning `geoblocked = true` if we're not force allow connecting and they're
    // geoblocked.
    return !forceAllowConnect && shouldBeGeoblocked;
  }, [forceAllowConnect, shouldBeGeoblocked]);
```

CRBl69 (approved):
Small readability change request. Once accepted, and you checked that it works, feel free to merge.

PR #341, opened 2024-11-12
[ECO-2364] Fix double rendering issue on home page

CRBl69:
# Description

This PR fixes a bug where the home page grid would double render when searching for something.

# Testing

Search, check no double animation.

# Checklist


xbtmatt (approved):

xbtmatt (approved):
Note that this change works because when the user was sorting by bump order + searching for emojis it would still animate, but that was eventually disabled, and thus we no longer need the searchEmojis in the component key. This removes the extra re-render.

PR #364, opened 2024-11-17
[ECO-2429] Move the `emoji => name` middleware conversion to be before the allow…

xbtmatt:
# Description

Right now, if the allowlist is disabled, `middleware.ts` will call `NextResponse.next()` before normalizing the market path (converting emojis to their corresponding text names) and thus causing the page to return emoji not found.

- [x] To fix this, simply put the normalization of the market path before the allowlist check.

This will still require the user to verify once they're redirected to the normalized market path version of their URL, but will not result in the emoji not found error.

NOTE: This was actually in two places, so it looks like I'm removing the check entirely, but there's another identical check right after the path normalization.

# Testing

this PR.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/middleware.ts
+++ b/src/typescript/frontend/src/middleware.ts
@@ -21,10 +21,9 @@ export default async function middleware(request: NextRequest) {
     return NextResponse.next();
   }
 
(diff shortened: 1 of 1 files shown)


CRBl69 (approved):
LGTM AMRW LGTPRI OYMF

PR #390, opened 2024-11-20
[ECO-2460] Implement server side api key

CRBl69:
# Description

This PR adds support for a server side API key.


xbtmatt (changes requested):
I realized shortly after you went to bed that the best way ato do this is probably to nullish coalesce a server env and a client env, because the server env will always be undefined on the client. it will default to the client env if the server env is undefined.

And we can ensure that the server env is defined on the server by throwing an error if it's not.

basically, `API_KEY = SERVER_API_KEY ?? CLIENT_API_KEY;` and in a build file that's on the server:

```javascript
"use server";
if (!SERVER_API_KEY) {
    throw new Error("Server env is undefined");
}
```

just rough code but you get the idea

src/typescript/example.env
```diff
@@ -68,3 +68,15 @@ PUBLISHER_PRIVATE_KEY=""
...
+NEXT_PUBLIC_LOCAL_APTOS_API_KEY=""
+NEXT_PUBLIC_CUSTOM_APTOS_API_KEY=""
+NEXT_PUBLIC_DEVNET_APTOS_API_KEY=""
+NEXT_PUBLIC_TESTNET_APTOS_API_KEY=""
+NEXT_PUBLIC_MAINNET_APTOS_API_KEY=""
+SERVER_LOCAL_APTOS_API_KEY=""
+SERVER_CUSTOM_APTOS_API_KEY=""
+SERVER_DEVNET_APTOS_API_KEY=""
+SERVER_TESTNET_APTOS_API_KEY=""
+SERVER_MAINNET_APTOS_API_KEY=""
```
xbtmatt, line 82:
```suggestion
# These should NOT be exposed publicly.
SERVER_LOCAL_APTOS_API_KEY=""
SERVER_CUSTOM_APTOS_API_KEY=""
SERVER_DEVNET_APTOS_API_KEY=""
SERVER_TESTNET_APTOS_API_KEY=""
SERVER_MAINNET_APTOS_API_KEY=""
```

xbtmatt (approved):

PR #392, opened 2024-11-20
[ECO-2463] Separate the actual txn simulation for gas usage calculation from the swap output calculation

CRBl69:
# Description

This PR does the thing.


src/typescript/frontend/src/lib/hooks/queries/use-simulate-swap.ts
```diff
@@ -98,7 +119,7 @@ export const useSimulateSwap = (args: {
       marketAddress,
       inputAmount.toString(),
       isSell,
-      numSwaps,
+      Math.round(numSwaps / 10) * 10,
```
xbtmatt, line 122:
```suggestion
      Math.floor(numSwaps / 10) * 10,
```

xbtmatt (approved):

PR #399, opened 2024-11-20
[ECO-2470] Add okx icon/logo to middleware.ts so it shows up in the wallet adapter on `/verify`

xbtmatt:
# Description

Fairly self explanatory- the image won't load unless the matcher allows it through on `middleware.ts` unauthenticated (that is, let an unauthenticated user see the okx logo in the wallet adapter).

We didn't have to do this before because the wallet adapter used all SVGs and those aren't public/static files, but pure HTML elements.

### DIFF
--- a/src/typescript/frontend/src/middleware.ts
+++ b/src/typescript/frontend/src/middleware.ts
@@ -20,14 +20,7 @@ export default async function middleware(request: NextRequest) {
   if (MAINTENANCE_MODE && pathname !== "/maintenance") {
     return NextResponse.redirect(new URL(ROUTES.maintenance, request.url));
   }
-  if (
-    pathname === "/social-preview.png" ||
-    pathname === "/webclip.png" ||
-    pathname === "/icon.png" ||
-    pathname === "/test" ||
-    pathname === "/geolocation" ||
-    pathname === "/verify_status"
-  ) {
+  if (pathname === "/test" || pathname === "/verify_status") {
     return NextResponse.next();
   }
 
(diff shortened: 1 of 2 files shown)


alnoki (approved):

src/typescript/frontend/src/middleware.ts
```diff
@@ -64,5 +64,5 @@ export default async function middleware(request: NextRequest) {
 }
 
 export const config = {
-  matcher: "/((?!verify|api|_next/static|_next/image|favicon.ico|logo192.png|manifest.json).*)",
+  matcher: "/((?!verify|api|_next/static|_next/image|favicon.ico|logo192.png|okx-logo.png|manifest.json).*)",
```
CRBl69, line 67: Just read the docs on this. Do you mind also adding items from this:

```
if (
    pathname === "/social-preview.png" ||
    pathname === "/webclip.png" ||
    pathname === "/icon.png" ||
    pathname === "/test" ||
    pathname === "/geolocation" ||
    pathname === "/verify_status"
  ) {
    return NextResponse.next();
}
```

I think that `*.png` could be added. Also, is `geolocation` needed anymore ?

CRBl69 (changes requested):

src/typescript/frontend/src/middleware.ts
xbtmatt: Tried to resolve this in: a92c191433dd508a33c873a448092c95b255727b

Will need to check manually (I'll check on preview build that middleware isn't running on /test)

CRBl69 (approved):

PR #402, opened 2024-11-21
[ECO-2483] Cache VPN queries

CRBl69:
# Description

Cache VPN queries using local storage.

### DIFF
--- a/src/typescript/frontend/src/configs/local-storage-keys.ts
+++ b/src/typescript/frontend/src/configs/local-storage-keys.ts
@@ -27,8 +27,12 @@ export function readLocalStorageCache<T>(key: keyof typeof LOCAL_STORAGE_KEYS): 
     return null;
   }
   const cache = parseJSON<LocalStorageCache<T>>(str);
-  if (new Date(cache.expiry) > new Date()) {
-    return cache.data;
+  try {
+    if (new Date(cache.expiry) > new Date()) {
+      return cache.data;
+    }
+  } catch (e) {
+    return null;
   }
   return null;
 }
--- a/src/typescript/frontend/src/configs/local-storage-keys.ts
+++ b/src/typescript/frontend/src/configs/local-storage-keys.ts
@@ -19,7 +19,9 @@ export type LocalStorageCache<T> = {
 };
 
 /**
- * Note that this data is not validated and any change in data type returned from this function should be validated to ensure that persisted cache data between multiple builds can cause errors with unexpected data types.
+ * Note that this data is not validated and any change in data type returned from this function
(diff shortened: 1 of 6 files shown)


src/typescript/frontend/src/configs/local-storage-keys.ts
```diff
@@ -1,8 +1,39 @@
...
+  language: Infinity,
+  geoblocking: 7 * 24 * 60 * 60 * 1000, // 7 days.
+};
+
+export type LocalStorageCache<T> = {
+  expiry: number;
+  data: T | null;
+};
+
+export function readLocalStorageCache<T>(key: keyof typeof LOCAL_STORAGE_KEYS): T | null {
```
xbtmatt, line 21:
```suggestion
/**
 * Note that this data is not validated and any change in data type returned from this function should be validated to ensure that persisted cache data between multiple builds can cause errors with unexpected data types.
 */
export function readLocalStorageCache<T>(key: keyof typeof LOCAL_STORAGE_KEYS): T | null {
```

src/typescript/frontend/src/configs/local-storage-keys.ts
```diff
@@ -1,8 +1,41 @@
...
+  language: Infinity,
+  geoblocking: 7 * 24 * 60 * 60 * 1000, // 7 days.
+};
+
+export type LocalStorageCache<T> = {
+  expiry: number;
+  data: T | null;
+};
+/**
+ * Note that this data is not validated and any change in data type returned from this function should be validated to ensure that persisted cache data between multiple builds can cause errors with unexpected data types.
```
xbtmatt, line 21:
```suggestion
};

/**
 * Note that this data is not validated and any change in data type returned from this function should be validated to ensure that persisted cache data between multiple builds can cause errors with unexpected data types.
```

src/typescript/frontend/src/configs/local-storage-keys.ts
```diff
@@ -1,8 +1,42 @@
...
+ */
+export function readLocalStorageCache<T>(key: keyof typeof LOCAL_STORAGE_KEYS): T | null {
+  const str = localStorage.getItem(key);
+  if (str === null) {
+    return null;
+  }
+  const cache = parseJSON<LocalStorageCache<T>>(str);
+  if (new Date(cache.expiry) > new Date()) {
+    return cache.data;
+  }
```
xbtmatt, line 32:
```suggestion
  const cache = parseJSON<LocalStorageCache<T>>(str);
  try {
    if (new Date(cache.expiry) > new Date()) {
      return cache.data;
    }
  } catch (e) {
    return null;
  }
```

xbtmatt (approved):

PR #409, opened 2024-11-22
[ECO-2493] Change chart to use quote volume, set default visible range

CRBl69:
# Description

- Change chart to use quote volume
- Set the chart's default visible range
- Update the number of swaps/chats delivered on load (`100` => `25`)
- Change default timeframe to 1m from 5m


src/typescript/frontend/src/app/chats/route.ts
```diff
@@ -0,0 +1,43 @@
...
+const isValidChatSearchParams = (params: ChatSearchParams): params is ValidChatSearchParams => {
+  const { marketID, fromMarketNonce } = params;
+  // prettier-ignore
+  return (
+    marketID !== null && isNumber(marketID) &&
+    fromMarketNonce !== null && isNumber(fromMarketNonce)
+  );
+};
+
+export async function GET(request: NextRequest) {
```
xbtmatt, line 26: I realize after trying to help Greg with our /api routes, we should probably add a simple example as a jsdoc string above these

You can probably write one up really quick with ChatGPT or something, should be pretty straightforward, just ask it to write a jsdoc but give it the context on the search params and any caveats, etc, then tell it to write an example

src/typescript/frontend/src/app/chats/route.ts
```diff
@@ -0,0 +1,43 @@
...
+  marketID: string | null;
+  fromMarketNonce: string | null;
+};
+
+export type ValidChatSearchParams = {
+  marketID: string;
+  fromMarketNonce: string;
+};
+
+const isNumber = (s: string) => !isNaN(parseInt(s));
```
xbtmatt, line 15:
```suggestion
const isNumber = (s: string) => {
  try {
    return !isNaN(parseInt(s));
  } catch {
    return false;
  }
}
```

Should always wrap these with a try/catch since it's coming from a search param and a malformed request somewhere will throw an error

src/typescript/frontend/src/app/trades/route.ts
```diff
@@ -0,0 +1,43 @@
...
+  marketID: string | null;
+  fromMarketNonce: string | null;
+};
+
+export type ValidSwapSearchParams = {
+  marketID: string;
+  fromMarketNonce: string;
+};
+
+const isNumber = (s: string) => !isNaN(parseInt(s));
```
xbtmatt, line 15: Let's not repeat this, use the other isNumber, and find the other one we use (I think I wrote a similar one somewhere, maybe like safeParseInt or something, it was for another route/fetcher/getter)

src/typescript/frontend/src/app/trades/route.ts
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  if (!isValidSwapSearchParams(params)) {
+    return new Response("Invalid chat search params.", { status: 400 });
+  }
+
+  const marketID = Number(params.marketID);
+  const fromMarketNonce = Number(params.fromMarketNonce);
+
+  const res = await fetchSwapEvents({ marketID, fromMarketNonce });
```
xbtmatt, line 40: We should probably cache these like we do with candlestick data- although we might want to strip most of the data first (since I think we quite literally only use like 3-4 fields from this endpoint), but it's a lot better than hitting the indexer

It should also be simpler to cache these compared to candlesticks

src/typescript/frontend/src/app/chats/route.ts
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  if (!isValidChatSearchParams(params)) {
+    return new Response("Invalid chat search params.", { status: 400 });
+  }
+
+  const marketID = Number(params.marketID);
+  const fromMarketNonce = Number(params.fromMarketNonce);
+
+  const res = await fetchChatEvents({ marketID, fromMarketNonce });
```
xbtmatt, line 40: We should cache these as well (like the swap events)

xbtmatt (changes requested):
1. Remove unused fields from swap/chat events when fetching from the indexer (based on what the market page uses, it's like 4-5 fields total, you can get rid of most of the market state/periodic state event stuff)
2. Cache the swap/chat data like we do with candlesticks
3. Consolidate the safe number parsing logic into a single function with a try/catch that returns false if it throws an error

src/typescript/frontend/src/app/chats/route.ts
CRBl69: I'll note that `parseInt("lol")` just returns `NaN` so I don't think this is a problem.

xbtmatt (commented):
Discussed on eng call: split this PR up into two PRs where the first one is simply the more basic performance optimizations with infinite load, and the second is the generic, parameterizable `Parcelize` functionality that will be a robust solution we can always use in the future.

And then we can properly unit test and abstract the Parcelize logic without having to rush to get it in

Also you can address this in the later PR too: 

```
I realize after trying to help Greg with our /api routes,
we should probably add a simple example as a jsdoc string above all of our GET routes/requests

You can probably write one up really quick with ChatGPT or something,
should be pretty straightforward,
just ask it to write a jsdoc but give it the context on the search params and any caveats, etc, then tell it to write an example
```

CRBl69:
Noting that all parcel changes were moved to #435 in order to merge this ASAP.

xbtmatt (approved):

PR #410, opened 2024-11-22
[ECO-2496] Incorporate misc changes from the payment solution PRs

CRBl69:
# Description

This PR adds:

- save settings to local storage
- have a global layer spacing instead of specifing a padding on every page
- load emojis from emoji-mart lib in providers so that it's loaded on every page

### DIFF
--- a/src/typescript/frontend/src/configs/local-storage-keys.ts
+++ b/src/typescript/frontend/src/configs/local-storage-keys.ts
@@ -5,12 +5,14 @@ const LOCAL_STORAGE_KEYS = {
   theme: `${packages.name}_theme`,
   language: `${packages.name}_language`,
   geoblocking: `${packages.name}_geoblocking`,
+  settings: `${packages.name}_geoblocking`,
 };
 
 const LOCAL_STORAGE_CACHE_TIME = {
   theme: Infinity,
   language: Infinity,
   geoblocking: 7 * 24 * 60 * 60 * 1000, // 7 days.
+  settings: Infinity, // 7 days.
 };
 
 export type LocalStorageCache<T> = {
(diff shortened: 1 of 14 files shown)


src/typescript/frontend/src/configs/local-storage-keys.ts
```diff
@@ -5,12 +5,14 @@ const LOCAL_STORAGE_KEYS = {
   theme: `${packages.name}_theme`,
   language: `${packages.name}_language`,
   geoblocking: `${packages.name}_geoblocking`,
+  settings: `${packages.name}_geoblocking`,
```
xbtmatt, line 8:
```suggestion
  settings: `${packages.name}_settings`,
```

src/typescript/frontend/src/configs/local-storage-keys.ts
```diff
@@ -5,12 +5,14 @@ const LOCAL_STORAGE_KEYS = {
...
 };
 
 const LOCAL_STORAGE_CACHE_TIME = {
   theme: Infinity,
   language: Infinity,
   geoblocking: 7 * 24 * 60 * 60 * 1000, // 7 days.
+  settings: Infinity, // 7 days.
```
xbtmatt, line 15:
```suggestion
  settings: Infinity,
```

xbtmatt (changes requested):
I think you forgot to do the [market]/page.tsx page right? Don't see the changes in here and it looks a little funky. Think you covered the others though

Other than that looks good, would approve but don't want an accidental auto-merge

src/typescript/frontend/src/configs/local-storage-keys.ts
CRBl69: My keyboard:

image

xbtmatt (commented):
The padding at the top of the `[market]/page.tsx` is still a little different than before. Is that intended?

It looks okay, but I didn't check every window size to see if it's actually responsive with mobile and whatnot. Just want to make sure you changed it since I didn't see it in the components you updated.

alnoki (approved):

xbtmatt (approved):

PR #412, opened 2024-11-25
[ECO-2492] Calculate the swap price client-side

xbtmatt:
# Description

Instead of hitting the Aptos Labs fullnode with a view function, we can run the `simulate_swap_inner` function client-side. This will lower the resources we consume from the fullnode API.

- [x] Port the Move contract logic to TypeScript in the SDK
- [x] Add a hook function to wrap the calculation function with inputs from the frontend
- [x] Remove the `useSimulateSwap` hook (or only use it as a backup)
- [x] Remove any remaining console logs/manual testing code

# Testing

- [x] Verify `useSimulateSwap` always results in the same output as the calculation function
- [x] Test it on the preview build with a new market and an existing market, compare to results from the existing `production` build or another build with the old logic
- [x] e2e tests to verify the logic is correct in multiple circumstances

# Checklist


src/typescript/frontend/src/lib/hooks/queries/use-simulate-swap.ts
```diff
@@ -165,3 +168,40 @@ export const useSimulateSwap = (args: {
...
+    aptos,
+    account,
+    ...args,
+    swapper,
+    inputAmount,
+    minOutputAmount,
+    typeTags,
+  });
+
+  // Neither of these values will ever be zero, so we can just check if it's truthy.
```
xbtmatt, line 203:
```suggestion
  // Neither of these values will ever be zero if a meaningful value is returned,
  // so we can just check if it's truthy.
```

src/typescript/sdk/src/emojicoin_dot_fun/calculate-swap-price.ts
```diff
@@ -0,0 +1,216 @@
...
+  const denominator = inputAmount.plus(denominatorAddend.toString());
+  if (!denominator.gt(0)) {
+    throw new DivideByZeroError();
+  }
+  return numerator.div(denominator);
+};
+
+/**
+ * @throws {@link SwapNotEnoughBaseError} if the user does not have enough base
+ * @throws {@link DivideByZeroError} if the cpamm output calculation results in a divide by zero
```
xbtmatt, line 70:
```suggestion
 * @throws {@link DivideByZeroError} if the cpamm output calculation results in a divide by zero
 * @see {@link https://github.com/econia-labs/emojicoin-dot-fun/blob/295cf611950f66651452baa3e6ad6d6aef583f9b/src/move/emojicoin_dot_fun/sources/emojicoin_dot_fun.move#L1691}
```

alnoki (approved):

PR #417, opened 2024-11-27
[ECO-2501] Fix the WSOD with Aptos Connect adapter on Firefox/Safari

xbtmatt:
# Description

WSOD on Firefox/Safari when submitting txns. We need to unroll the wallet adapter transaction submission so there's no promises while the window tries to open a pop-up, or the browsers will say the application has crashed.

# Testing

Manually with the preview build. Also we need to test on mobile, too.

- [x] Ensure it works on mobile on safari
- [x] Ensure all state is still loading from various events
- [x] Things should load in swap history
- [x] Check nearly everything on the site where you can submit
- [x] Ensure that any buttons that submit transactions are disabled while the builder is loading (if props change, it should temporarily be disabled while the builder finishes its async call)

Extra things fixed/changed during this process:
- For anything but register market (anything that doesn't use the sign *then* submit flow), the transaction builder no longer makes async calls to the fullnode when building a transaction. The wallet adapter now handles it. This reduces the amount of API calls our frontend does and reduces the amount of time from click => transaction preview from ~150ms to ~5ms
- Sequence number is now fetched on an interval/on demand, for the register market flow. This also reduces the fullnode queries and speeds it up in general, even for sign then submit flow

# Checklist


CRBl69 (approved):
exquisite

PR #431, opened 2024-11-29
[ECO-2266] Add go to page button

CRBl69:
# Description

This PR adds a button to go to the market page from the pools page.

This also optimizes the pools page by using `Number` instead of `Big` for the XPR calculations. This looses some precision but is not that visible.

It also refactors the `Popup` component with default styling.


xbtmatt (approved):
Significantly faster on the XPR calculations, thanks for doing this! The new link looks great too

PR #432, opened 2024-11-30
[ECO-2529] Refactor the ANS name resolver to not cause infinite renders

xbtmatt:
# Description

The ANS name resolver was returning an anonymous function in the `NameStore` (name-store.ts). This meant that the value returned was causing a re-render.  Since the header displays the connected user's ANS name, this caused a downward effect of an unlimited number of re-renders- albeit in a somewhat slow and controlled manner on the home page. On the market page, the re-rendering back and forth between the trade history, chat box, and the user's address in the header meant that there was a near `n^3` re-render effect (where `n = numChats + numSwaps` ) thus causing massive amounts of re-renders when the number of chats and swaps became somewhat large.

- [x] Refactor the `nameResolver`
- [x] Have it use a map that resolves the name and returns the input address while it is either unresolved or not found
- [x] Move the name resolution into a hook to combine the resolve + map.get into one function based on the address
- [x] Move the resolution as deep into the component tree as possible
- [x] Add the name resolution for the user's address into the `AptosContextProvider` as `addressName`
- [x] Reduce the local storage cache time from 7 days to 1 day if the resolved name returns `null`, that way newly registered names will show up in 1 day rather than in 7.

# Testing

- [x] Test other user's names
- [x] Test connecting with a wallet with an ANS name
- [x] Test swapping/chatting with an ANS name
- [x] Test swapping/chatting with an ANS name and then switching wallets, make sure chat/trade history is updated
- [x] Then switch back, to make sure it still shows up as you as the sender (for the chat box)

# Checklist

### DIFF
--- a/src/typescript/frontend/src/lib/store/name-store.ts
+++ b/src/typescript/frontend/src/lib/store/name-store.ts
@@ -1,36 +1,31 @@
 import { createStore } from "zustand/vanilla";
 import { immer } from "zustand/middleware/immer";
 import { APTOS_NETWORK } from "lib/env";
-import type { WritableDraft } from "immer";
+import { MS_IN_ONE_DAY } from "components/charts/const";
 
 export type ANSValue = { name: string | null; expiry: number };
 export type ANSMap = Map<string, ANSValue>;
 export type ANSPromiseMap = Map<string, Promise<void>>;
-
 export type NameState = {
   names: ANSMap;
 };
 export type NameActions = {
   /**
-   * Return a map of addresses to names.
+   * Resolves an ANS name for an address. If it's not in local storage, it will be fetched.
    *
-   * The returned map might not contain the values right away, as it takes time to make network requests.
-   */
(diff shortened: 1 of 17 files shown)


src/typescript/frontend/src/lib/store/name-store.ts
```diff
@@ -118,54 +113,31 @@ export const initializeEventStore = (): NameState => {
...
 export const createNameStore = (initialState: NameState = defaultState) => {
   return createStore<NameStore>()(
     immer((set, get) => ({
       ...initialState,
-      getMapWithNames: (addresses) => {
-        return getNameMap(set, get, addresses);
-      },
-      getResolverWithNames: (addresses) => {
-        const map = getNameMap(set, get, addresses);
-        return (address) => map.get(address)?.name ?? address;
```
xbtmatt, line 168: For future reference @CRBl69, this was the offending line. Feel free to resolve once you see this

CRBl69 (approved):
Wow. Simply lovely. Thank you !

PR #433, opened 2024-11-30
[ECO-2530] Fix emojis not showing up on most devices; add `jest` setup for frontend

xbtmatt:
# Description

This means all smart-contract supported emojis will show up on the website for everyone now!

The client will deliver a fairly large font package now, but it's worth it for non-Apple devices. Apple devices get special treatment because the Apple emoji font is gorgeous and obviously lightweight for those users.

- [x] Add noto color emoji font for non apple users by checking user agent headers in the `layout.tsx` file
- [x] Remove the `em-emoji` element since it's somewhat heavy and needlessly increases the overall DOM size since we use *so* many emojis everywhere
- [x] Now we just use noto-color-emoji but it falls back to the others if they exist
- [x] Add `jest` setup for the frontend. Yay finally!
- [x] Fix the emoji loading circle not showing them (likely a css issue)
- [x] Fix the sizing of different emojis (like in wallet adapter lightning bolt connected emojis)
- [x] Change all emoji picker emojis to match the noto emojis
- [x] Use noto emojis in the noto picker textbox
- [x] Ensure the rocket emoji on market metadata looks correct
- [x] Remove console.logs from output

# Testing

~- [ ] Ensure users don't download the font if they're on an iOS/Mac device~

This is actually possible if we don't force a preload by including it in an element in the top `layout.tsx`, but it causes flashing on non-Apple devices, so it's better to include it and have NextJS preload the font.

- [x] Ensure the font is used if the user has it (Windows, Linux, various browsers, etc)
- [x] Make sure nothing else breaks

# Checklist


MoonShiesty:
> because the Apple emoji font is gorgeous and obviously lightweight for those users

i think its worth weighting whether its better to have a single canonical emoji image, that matches the FA uri, vs reducing the package size on devices with good native emoji fonts

> noto-color-emoji

i dont like the noto font, twitter's is far better

noto doesn't have emojis for many code points, ive noticed missing skin tones, flags etc

the noto US flag emoji looks like dogshit, as do many others. twemoji-15.0.3 font is open-source and much better

https://emojipedia.org/twitter/twemoji-15.0.3

image

image

xbtmatt:
> > because the Apple emoji font is gorgeous and obviously lightweight for those users
> 
> i think its worth weighting whether its better to have a single canonical emoji image, that matches the FA uri, vs reducing the package size on devices with good native emoji fonts
> 
> > noto-color-emoji
> 
> i dont like the noto font, twitter's is far better
> 
> noto doesn't have emojis for many code points, ive noticed missing skin tones, flags etc
> 
> the noto US flag emoji looks like dogshit, as do many others. twemoji-15.0.3 font is open-source and much better
> 
> https://emojipedia.org/twitter/twemoji-15.0.3
> 
> image image

Hey @MoonShiesty! Thanks for the comment, I think that's just because you don't have the Noto font installed..? They should have set up the site to download it for you but apparently it isn't working..?

When I use google's font SDK with nextjs it shows up just fine, see here: 
image

alnoki (approved):

PR #422, opened 2024-11-28
[WIP] Add endpoints for Dexscreener support

CapCap:
# Description

Delete this sentence and add a description that explains what changes you have
done and why they were necessary.

# Testing

Delete this sentence and provide a description of how to test the changes in
this PR.

# Checklist

- [ ] Did you update relevant documentation?
- [ ] Did you add tests to cover new code or a fixed issue?
- [ ] Did you update the changelog?
- [ ] Did you check all checkboxes from the linked Linear task?

PR #446, opened 2024-12-05
Add endpoints for Dexscreener support #422

CapCap:
Continuation of  #422 

Dexscreener API stubs

Dexscreener events cconverion v1

Dexscreener events cconverion v1

refac out of test

Scope out other enddex points

small lints

Event feedback changes

API mapping updates

Add block and event index to API

set fee bps

use reserve calculation from SDK

decimalize outputs

# Description

Delete this sentence and add a description that explains what changes you have
done and why they were necessary.

# Testing

this PR.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/app/dexscreener/events.ts
+++ b/src/typescript/frontend/src/app/dexscreener/events.ts
@@ -83,8 +83,7 @@ import {
   type toLiquidityEventModel,
   type toSwapEventModel,
 } from "@sdk/indexer-v2/types";
-import Big from "big.js";
-import { calculateRealReserves } from "@sdk/markets";
+import { calculateCurvePrice, calculateRealReserves } from "@sdk/markets";
 import { toCoinDecimalString } from "../../lib/utils/decimals";
 import { DECIMALS } from "@sdk/const";
 import { symbolEmojisToPairId } from "./util";
@@ -182,8 +181,6 @@ export interface EventsResponse {
   events: Event[];
 }
 
-const TWO_TO_64 = new Big(2).pow(64);
-
 export function toDexscreenerSwapEvent(
   event: ReturnType<typeof toSwapEventModel>
 ): SwapEvent & BlockInfo {
@@ -213,9 +210,7 @@ export function toDexscreenerSwapEvent(
     asset1: toCoinDecimalString(quote, DECIMALS),
   };
 
-  const priceNative = new Big(event.swap.avgExecutionPriceQ64.toString())
-    .div(TWO_TO_64)
-    .toFixed(64);
+  const priceNative = calculateCurvePrice(event.state).toFixed(50).toString();
 
   if (!event.blockAndEvent) throw new Error("blockAndEvent is undefined");
 
--- a/src/typescript/frontend/src/app/dexscreener/events.ts
+++ b/src/typescript/frontend/src/app/dexscreener/events.ts
@@ -1,0 +1,308 @@
+/***
+ Request: GET /events?fromBlock=:number&toBlock=:number
+
+ - fromBlock and toBlock are both inclusive: a request to /events?fromBlock=10&toBlock=15 should include all
+ available events from block 10, 11, 12, 13, 14 and 15
+
+ Response Schema:
+ // interface EventsResponse {
+ //  events: Array<{ block: Block } & (SwapEvent | JoinExitEvent)>;
+ // }
+
+ Example Response:
+ // {
+ //   "events": [
+ //     {
+ //       "block": {
+ //         "blockNumber": 10,
+ //         "blockTimestamp": 1673319600
+ //       },
+ //       "eventType": "swap",
+ //       "txnId": "0xe9e91f1ee4b56c0df2e9f06c2b8c27c6076195a88a7b8537ba8313d80e6f124e",
+ //       "txnIndex": 4,
+ //       "eventIndex": 3,
+ //       "maker": "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045",
+ //       "pairId": "0x123",
+ //       "asset0In": 10000,
+ //       "asset1Out": 20000,
+ //       "priceNative": 2,
+ //       "reserves": {
+ //         "asset0": 500,
+ //         "asset1": 1000
+ //       }
+ //     },
+ //     {
+ //       "block": {
+ //         "blockNumber": 10,
+ //         "blockTimestamp": 1673319600
+ //       },
+ //       "eventType": "join",
+ //       "txnId": "0xea1093d492a1dcb1bef708f771a99a96ff05dcab81ca76c31940300177fcf49f",
+ //       "txnIndex": 0,
+ //       "eventIndex": 0,
+ //       "maker": "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045",
+ //       "pairId": "0x456",
+ //       "amount0": 10,
+ //       "amount1": 5,
+ //       "reserves": {
+ //         "asset0": 100,
+ //         "asset1": 50
+ //       }
+ //     },
+ //     {
+ //       "block": {
+ //         "blockNumber": 11,
+ //         "blockTimestamp": 1673406000
+ //       },
+ //       "eventType": "swap",
+ //       "txnId": "0xea1093d492a1dcb1bef708f771a99a96ff05dcab81ca76c31940300177fcf49f",
+ //       "txnIndex": 1,
+ //       "eventIndex": 20,
+ //       "maker": "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045",
+ //       "pairId": "0x456",
+ //       "asset0In": 0.0123456789,
+ //       "asset1Out": 0.000123456789,
+ //       "priceNative": 0.00000012345,
+ //       "reserves": {
+ //         "asset0": 0.0001,
+ //         "asset1": 0.000000000000001
+ //       }
+ //     }
+ //   ]
+ // }
+ **/
+
+import { type NextRequest, NextResponse } from "next/server";
+import { type Block } from "./latest-block";
+import {
+  fetchLiquidityEventsByBlock,
+  fetchSwapEventsByBlock,
+} from "@sdk/indexer-v2/queries/app/dexscreener";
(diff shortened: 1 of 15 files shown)


src/typescript/sdk/src/indexer-v2/types/index.ts
```diff
@@ -51,6 +51,11 @@ const toTransactionMetadata = (
   insertedAt: data.inserted_at ? postgresTimestampToDate(data.inserted_at) : new Date(0),
 });
 
+const toBlockAndEventIndex = (data: BlockAndEventIndexMetadata) => ({
+  blockNumber: BigInt(data.block_number),
+  eventIndex: BigInt(data.event_index),
+});
+
```
xbtmatt, line 58:
```suggestion
const toBlockAndEventIndex = (data: BlockAndEventIndexMetadata) => data.block_number ? {
  blockNumber: BigInt(data.block_number),
  eventIndex: BigInt(data.event_index),
} : undefined;

```

You should just be able to have this return `undefined` if `block_number` is undefined, and then you don't have to change anything with the `mini-processor` stuff, since the entire `blockAndEvent` field will be undefined, rather than the individual fields

It will work with `curryToNamedType` as expected

xbtmatt: Yeah just checked!

```typescript
const toBlockAndEventIndex = (data: BlockAndEventIndexMetadata) => ({
  blockNumber: BigInt(data.block_number),
  eventIndex: BigInt(data.event_index),
});

```
const possiblyWithData = curryToNamedType(
  (argsIn: number) => argsIn === 2 ? { data: argsIn } : undefined,
  "maybeData"
);

const res = possiblyWithData(1);
```

and then typescript resolves the type to: 

```typescript
const res: {
    maybeData: {
        data: number;
    } | undefined;
}
```

src/typescript/sdk/src/indexer-v2/types/json-types.ts
```diff
@@ -125,6 +125,11 @@ type TransactionMetadata = {
   inserted_at?: PostgresTimestamp;
 };
 
+export type BlockAndEventIndexMetadata = {
+  block_number: Uint64String,
+  event_index: number,
```
xbtmatt, line 130:
```suggestion
  block_number?: Uint64String,
  event_index?: number,
```

And then you'd do this, because the data is undefined at the individual field level here since it's the json type

xbtmatt (commented):
Just wanted to quickly review the block and event index stuff- it doesn't have to be implemented the way I suggested in the comments if you have a better idea, but I think it's simple enough

src/typescript/frontend/src/app/dexscreener/events.ts
```diff
@@ -0,0 +1,314 @@
...
+    };
+  }
+
+  const { base, quote } = calculateRealReserves(event.state);
+  const reserves = {
+    asset0: toCoinDecimalString(base, DECIMALS),
+    asset1: toCoinDecimalString(quote, DECIMALS),
+  };
+
+  const priceNative = new Big(event.swap.avgExecutionPriceQ64.toString()).div(2 ** 64).toFixed(64);
```
xbtmatt, line 209:
```suggestion
  const priceNative = new Big(event.swap.avgExecutionPriceQ64.toString()).div(2n ** 64n).toFixed(64);
```

I think although you may not run into a loss of precision immediately here, this value is bigger than `Number.MAX_SAFE_INTEGER`, so we should just cast to bigint to be safe

src/typescript/frontend/src/app/dexscreener/pair.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/**
+ *
+ * @param pairId is the pair ID. Generally it's `event.market.symbolEmojis.join("") + "-APT"`
+ */
+export async function getPair(pairId: string): Promise<Pair> {
+  const sybolEmojis = pairIdToSymbolEmojis(pairId);
+
+  const marketRegistrations = await fetchMarketRegistrationEventBySymbolEmojis({ sybolEmojis });
```
xbtmatt, line 80:
```suggestion
  const symbolEmojis = pairIdToSymbolEmojis(pairId);

  const marketRegistrations = await fetchMarketRegistrationEventBySymbolEmojis({ symbolEmojis });
```

Though sybol emojis sounds interesting...

src/typescript/frontend/src/app/dexscreener/pair.ts
```diff
@@ -0,0 +1,109 @@
...
+  const marketRegistration = marketRegistrations[0];
```
xbtmatt, line 81: Maybe not necessary, but I think this could technically be undefined, may want to use `.pop()` here or check for `undefined` explicitly and return a `400` error if so

src/typescript/frontend/src/middleware.ts
```diff
@@ -20,7 +20,7 @@ export default async function middleware(request: NextRequest) {
   if (MAINTENANCE_MODE && pathname !== "/maintenance") {
     return NextResponse.redirect(new URL(ROUTES.maintenance, request.url));
   }
-  if (pathname === "/test" || pathname === "/verify_status") {
+  if (pathname === "/test" || pathname === "/verify_status" || pathname === "/dexscreener") {
```
xbtmatt, line 23:
```suggestion
  if (pathname === "/test" || pathname === "/verify_status" || pathname === ROUTES.dexscreener) {
```

I think you added it below, right? Although we probably should have added the others in there too... 😅

xbtmatt (approved):
This looks good, we'll just have to get the indexer stuff in before this (or the CI check for processor submodule will fail)

And then I think a `pnpm run format && pnpm run lint:fix` will fix any formatting/linter errors

src/typescript/frontend/src/app/dexscreener/events.ts
```diff
@@ -0,0 +1,314 @@
...
+): SwapEvent & BlockInfo {
+  let assetInOut;
+
+  if (event.swap.isSell) {
+    // We are selling to APT
+    assetInOut = {
+      asset0In: toCoinDecimalString(event.swap.inputAmount, DECIMALS),
+      asset0Out: 0,
+      asset1In: 0,
+      asset1Out: toCoinDecimalString(event.swap.quoteVolume, DECIMALS),
```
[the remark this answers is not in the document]
xbtmatt: Same with the other one too (with assert0/1 inverted)

src/typescript/frontend/src/app/dexscreener/events.ts
```diff
@@ -0,0 +1,308 @@
...
+    };
+  } else {
+    // We are buying with APT
+    assetInOut = {
+      asset0In: 0,
+      asset0Out: toCoinDecimalString(event.swap.quoteVolume, DECIMALS),
+      asset1In: toCoinDecimalString(event.swap.inputAmount, DECIMALS),
+      asset1Out: 0,
+    };
+  }
```
xbtmatt, line 208:
```suggestion
    assetInOut = {
      asset0In: toCoinDecimalString(event.swap.baseVolume, DECIMALS),
      asset0Out: 0,
      asset1In: 0,
      asset1Out: toCoinDecimalString(event.swap.quoteVolume, DECIMALS),
    };
  } else {
    // We are buying with APT
    assetInOut = {
      asset0In: 0,
      asset0Out: toCoinDecimalString(event.swap.baseVolume, DECIMALS),
      asset1In: toCoinDecimalString(event.swap.quoteVolume, DECIMALS),
      asset1Out: 0,
    };
  }
```
What you'd want to do is, in case of a `sell` and a `buy`:

`sell` means they're exchanging their emojicoins for APT (`base` for `quote`)
`buy` means they're exchanging APT for emojicoins (`quote` for `base`)

so if we treat `asset0/asset1` as `base/quote` respectively:

`sell`, `base` in `quote` out
 - `asset0in` (`base` in) would be `baseVolume`
 - `asset1out` (`quote` out) would be `quoteVolume`
`buy`, `quote` in, `base` out
 - `asset0out` (`base` out) would be `baseVolume`
 - `asset1In` (`quote` in) would be `quoteVolume`

which I think is quite different from the picture, now that I write it out hahaha. I believe I have it correct (in terms of in/out ) but regardless, I'm positive the sell = base for quote and buy = quote for base logic is 100% true

src/typescript/frontend/src/app/dexscreener/events.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  const { base, quote } = calculateRealReserves(event.state);
+  const reserves = {
+    asset0: toCoinDecimalString(base, DECIMALS),
+    asset1: toCoinDecimalString(quote, DECIMALS),
+  };
+
+  const priceNative = new Big(event.swap.avgExecutionPriceQ64.toString())
+    .div(TWO_TO_64)
+    .toFixed(64);
```
xbtmatt, line 218:
```suggestion
  const priceNative = calculateCurvePrice(event.state).toFixed(50).toString();
```

note the 50 decimal places here, because your comments on their API says to use as many as possible up to 50 (we use 100)

xbtmatt (commented):
Noted how to use `calculateCurvePrice` to get the value of `priceNative`

Main thing to note is that it always returns the price in `quote / base` aka `APT / emojicoin`, which is what the comments you wrote seem to suggest Dexscreener expects...but just noting here that it doesn't change.

If `asset0` and `asset1` are always `base / quote` respectively, then this should be fine

src/typescript/frontend/src/app/dexscreener/events.ts
CapCap: we still need the `.div(TWO_TO_64)` right?

xbtmatt: We wouldn't need it there - the reason we do the division by 2^64 on the average execution price is because it's a q64, which is essentially just a whole number representation of a fraction on-chain.

We avoid the usage of the q64 representation (and thus having to divide) by simply calculating the price client-side with the reserves amounts with `big.js`. It's technically more accurate and a little simpler even

xbtmatt (commented):
Should be good to go now format/lint-wise, the only thing I noticed is that the types in the notion doc are actually different from the types in the dexscreener website API.

Specifically one thing is the lack of `priceNative` in the dexscreener repo API, but it's in the notion doc.

Is the notion doc more accurate/a more specific version you made with them? Just want to clarify before merging

PR #448, opened 2024-12-06
[ECO-2525] Sort pool page by dpr by default

CRBl69:
# Description

This PR makes it so that by default, the pool page is sorted by APR.

# Testing

See pool page on vercel.

# Checklist


alnoki (approved):

xbtmatt (approved):

xbtmatt (approved):
I fixed this so it's the same on the server component as well- otherwise it flashes with the market cap sort first then apr

PR #471, opened 2024-12-19
[ECO-2612] Fix geoblocking logic

CRBl69:
# Description

## Geoblocking

This PR updates the geoblocking logic to handle cases where either the country or the region is not set.

| Case | Action taken |
| ---- | ------------ |
| country === `undefined` && region === `undefined` | geoblocked |
| country === `undefined` && region !== `undefined` | geoblocked if the region is restricted or the country of the region is restricted |
| country !== `undefined` && region === `undefined` | geoblocked if the country is restricted or a region in that country is restricted |
| country !== `undefined` && region !== `undefined` | geoblocked if the country is restricted or the region is restricted |

## Cache versionning

Additionally, all cache is now versionned using semantic versionning.

All previously present cache is assumed to be `1.0.0`.

When a major version changes, the cache is invalidated.

# Testing

Use VPN and connect to the website.

Check that the geoblocking is not cached improperly from previous version (you'll have to set it manually in local storage).

# Checklist

### DIFF
--- a/src/typescript/frontend/src/configs/local-storage-keys.ts
+++ b/src/typescript/frontend/src/configs/local-storage-keys.ts
@@ -1,7 +1,7 @@
 import { parseJSON, stringifyJSON } from "utils";
 import packages from "../../package.json";
 import { MS_IN_ONE_DAY } from "components/charts/const";
-import { satisfies, SemVer, parse } from "semver";
+import { satisfies, type SemVer, parse } from "semver";
 
 const LOCAL_STORAGE_KEYS = {
   theme: `${packages.name}_theme`,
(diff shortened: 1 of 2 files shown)


alnoki (approved):

src/typescript/frontend/src/configs/local-storage-keys.ts
```diff
@@ -19,20 +31,26 @@ export const LOCAL_STORAGE_CACHE_TIME = {
...
 export function readLocalStorageCache<T>(key: keyof typeof LOCAL_STORAGE_KEYS): T | null {
   const str = localStorage.getItem(LOCAL_STORAGE_KEYS[key]);
   if (str === null) {
     return null;
   }
   try {
     const cache = parseJSON<LocalStorageCache<T>>(str);
+    const range = `~${LOCAL_STORAGE_VERSIONS[key].major}`;
+    // Check for no breaking changes.
+    if (!satisfies(cache.version ?? "1.0.0", range)) {
```
xbtmatt, line 46: The only thing here I'd consider is whether or not this could accidentally default to 1.0.0 unintentionally- say, we forget to add the cache down the road. Since it defaults to 1.0.0, I mean.

But it's not enough to block this, so we can consider it later

src/typescript/frontend/src/utils/geolocation.ts
```diff
@@ -4,17 +4,29 @@ import { GEOBLOCKED, GEOBLOCKING_ENABLED } from "lib/server-env";
...
+      return true;
+    }
+  }
+  if (countryCode && !regionCode) {
+    if (GEOBLOCKED.regions.map((r) => r.split("-")[0]).includes(countryCode)) {
+      return true;
+    }
+  }
+  if (!countryCode && regionCode) {
+    if (GEOBLOCKED.countries.includes(regionCode.split("-")[0])) {
```
xbtmatt, line 27:
```suggestion
  if (!countryCode && regionCode) {
    // Note that even if the `regionCode` is `XX`, and `XX` is a banned country, this will return
    // `true` and thus block the user, because "XX".split("-")[0] is just "XX".
    if (GEOBLOCKED.countries.includes(regionCode.split("-")[0])) {
```

xbtmatt (approved):
The only thing here I'd consider is whether or not this could accidentally default to 1.0.0 unintentionally- say, we forget to add the cache down the road. Since it defaults to 1.0.0, I mean.

But it's not enough to block this, so we can consider it later

Overall this is great- appreciate you doing semantic versioning on these, that was definitely the ideal implementation :) Tested it live with different examples and it worked as expected. Added a note about splitting on a character not in the string, but other than that this looks good.

PR #478, opened 2024-12-20
[ECO-2618] Finalize emojicoin arena core logic design

alnoki:
# Description

This PR finalizes the Emojicoin Arena Move prototype, adding a README with a
comprehensive design overview.

Per extensive review, it adopts a lightweight data model that provides only the
minimal amount of information to required to enable comprehensive indexing,
as discussed in detail in the README.

## Logic updates

1. Add assorted inner abstractions to ease the review process.
1. Eliminate `SmartTable` structs since they can be a DoS vector.
1. Remove fields, events that are not required for lightweight data model.
1. Require that if a user has locked in since their most recent to an empty
   escrow, they must select lock in on subsequent top-off operations.
1. Allow user to elect lock-in even if they do not ultimately get matched, to
   prevent unexpected abortions.
1. Add exchange rate calculations.
1. Reset available rewards for a melee when it is cranked out.
1. Use integers instead of aggregators since exchange rate calculations require
   queries against both markets and thus inhibit parallelism within a melee
   in the general case.
1. Refactor type checking logic since exchange rate calculations do type checks.
1. Add assorted abstractions, helpers.
1. Add view functions and return unpackers for all core structs.
1. Add test-only fixturing to enable comprehensive state/event mocking.

## Test code

1. Add assorted mocking assertion functions.
1. Add additional test coin factories for `init_module` happy path tests.
1. Add `init_module` happy path tests for assorted pseudo-random seeds.
1. Add admin function happy path and expected failure tests

## Housekeeping

1. Use term `active` instead of `current` to indicate a melee that hasn't ended.
1. Update doc comments to link to actual structs, e.g. "`Escrow`" instead of
   "escrow".
1. Take out error codes used for admin-level guard rails that have since been
   removed.
1. Use compound assignment operators now that they are supported by `movefmt`
1. Change default melee time to 20 hours for new linear interpolation matching.

# Testing

From in `src/move/emojicoin_arena`:

```sh
git ls-files | entr -c sh -c " \
    aptos move compile --dev --move-2  &&
    aptos move fmt \
"
```

# Checklist

### DIFF
--- a/src/move/emojicoin_arena/sources/emojicoin_arena.move
+++ b/src/move/emojicoin_arena/sources/emojicoin_arena.move
@@ -31,14 +31,10 @@ module arena::emojicoin_arena {
     const E_INVALID_ESCROW_COIN_TYPE: u64 = 4;
     /// User has no escrow resource.
     const E_NO_ESCROW: u64 = 5;
-    /// Swapper has no funds in escrow to swap.
-    const E_SWAP_NO_FUNDS: u64 = 6;
     /// User has no funds in escrow to withdraw.
-    const E_EXIT_NO_FUNDS: u64 = 7;
+    const E_EXIT_NO_FUNDS: u64 = 6;
 
     const MAX_PERCENTAGE: u64 = 100;
-    const SHIFT_Q64: u8 = 64;
-    const NULL_ADDRESS: address = @0x0;
 
     /// Resource account address seed for the registry.
     const REGISTRY_SEED: vector<u8> = b"Arena registry";
@@ -180,48 +176,6 @@ module arena::emojicoin_arena {
         integrator_fee_rate_bps: u8
     }
 
-    public entry fun fund_vault(funder: &signer, amount: u64) acquires Registry {
-        let vault_address =
-            account::get_signer_capability_address(&Registry[@arena].signer_capability);
-        aptos_account::transfer(funder, vault_address, amount);
-        emit_vault_balance_update_with_vault_address(vault_address);
-    }
-
-    public entry fun set_next_melee_available_rewards(
-        arena: &signer, amount: u64
-    ) acquires Registry {
-        borrow_registry_ref_mut_checked(arena).next_melee_available_rewards = amount;
-    }
-
-    public entry fun set_next_melee_duration(arena: &signer, duration: u64) acquires Registry {
-        borrow_registry_ref_mut_checked(arena).next_melee_duration = duration;
-    }
-
-    public entry fun set_next_melee_max_match_percentage(
-        arena: &signer, max_match_percentage: u64
-    ) acquires Registry {
-        borrow_registry_ref_mut_checked(arena).next_melee_max_match_percentage =
-            max_match_percentage;
-    }
-
-    public entry fun set_next_melee_max_match_amount(
-        arena: &signer, max_match_amount: u64
-    ) acquires Registry {
-        borrow_registry_ref_mut_checked(arena).next_melee_max_match_amount =
-            max_match_amount;
-    }
-
-    public entry fun withdraw_from_vault(arena: &signer, amount: u64) acquires Registry {
-        let signer_capability_ref =
-            &borrow_registry_ref_mut_checked(arena).signer_capability;
-        aptos_account::transfer(
-            &account::create_signer_with_capability(signer_capability_ref),
-            @arena,
-            amount
-        );
-        emit_vault_balance_update_with_singer_capability_ref(signer_capability_ref);
-    }
-
     #[randomness]
     entry fun enter<Coin0, LP0, Coin1, LP1, EscrowCoin>(
         entrant: &signer, input_amount: u64, lock_in: bool
@@ -298,21 +252,6 @@ module arena::emojicoin_arena {
     }
 
     #[randomness]
-    entry fun exit<Coin0, LP0, Coin1, LP1>(participant: &signer) acquires Escrow, Registry {
-        // Crank schedule, set local variables.
-        let (registry_ref_mut, melee_is_active, participant_address, escrow_ref_mut) =
-            existing_participant_prologue<Coin0, LP0, Coin1, LP1>(participant);
-
-        exit_inner<Coin0, LP0, Coin1, LP1>(
-            registry_ref_mut,
-            escrow_ref_mut,
-            participant,
-            participant_address,
-            melee_is_active
-        );
-    }
-
-    #[randomness]
     entry fun swap<Coin0, LP0, Coin1, LP1>(swapper: &signer) acquires Escrow, Registry {
         // Crank schedule, set local variables.
         let (registry_ref_mut, melee_is_active, swapper_address, escrow_ref_mut) =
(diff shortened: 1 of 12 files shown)


src/move/emojicoin_arena/sources/emojicoin_arena.move
```diff
@@ -131,36 +125,169 @@ module arena::emojicoin_arena {
...
+        emojicoin_1_proceeds: u64
+    }
+
+    #[event]
+    /// Emitted whenever a user exits from `Escrow`.
+    struct Exit has copy, drop, store {
+        user: address,
+        melee_id: u64,
+        octas_entered: u64,
+        octas_matched: u64,
```
CRBl69, line 176: Could we keep the naming consistent with the event above ?

For example, I think `octas_matched` is basically `match_amount` and `octas_entered` is basically `input_amount`, yet they're named differently.

src/move/emojicoin_arena/sources/emojicoin_arena.move
```diff
@@ -131,36 +125,169 @@ module arena::emojicoin_arena {
...
+    /// Emitted after a user enters, swaps, or exits, representing their final `Escrow` state.
+    struct EscrowState has copy, drop, store {
+        user: address,
+        melee_id: u64,
+        emojicoin_0_balance: u64,
+        emojicoin_1_balance: u64,
+        n_swaps: u64,
+        swaps_volume: u128,
+        octas_entered: u64,
+        octas_matched: u64,
```
CRBl69, line 205: Same as above, can we keep this consistent ?

src/move/emojicoin_arena/sources/emojicoin_arena.move
```diff
@@ -131,36 +125,169 @@ module arena::emojicoin_arena {
...
+    }
+
+    #[event]
+    /// Emitted after a user enters, swaps, or exits, representing the final `Registry` state.
+    struct RegistryState has copy, drop, store {
+        n_melees: u64,
+        n_entrants: u64,
+        n_swaps: u64,
+        swaps_volume: u128,
+        octas_matched: u64,
```
CRBl69, line 243: Same consistency comment as above.

src/move/emojicoin_arena/sources/emojicoin_arena.move
```diff
@@ -131,36 +125,169 @@ module arena::emojicoin_arena {
...
+        top_exits: TopExits
+    }
+
+    #[event]
+    /// Emitted whenever a new `Melee` starts.
+    struct NewMelee has copy, drop, store {
+        melee_id: u64,
+        market_metadatas: vector<MarketMetadata>,
```
CRBl69, line 251: Is it possible to have `emojicoin_0_...` and `emojicoin_1_...` just like in the other places here ?

src/move/emojicoin_arena/sources/emojicoin_arena.move
```diff
@@ -131,36 +125,169 @@ module arena::emojicoin_arena {
...
+        start_time: u64,
+        duration: u64,
+        max_match_percentage: u64,
```
CRBl69, line 254: Shouldn't this be just `match_percentage` ? How would the match percentage be dynamic inside a melee (besides the last match(es) where there might not be enough APT left) ?

src/move/emojicoin_arena/sources/emojicoin_arena.move
```diff
@@ -131,36 +125,169 @@ module arena::emojicoin_arena {
...
+        available_rewards: u64
+    }
+
+    /// Exchange rate between APT and emojicoins.
+    struct ExchangeRate has copy, drop, store {
+        /// Octas per `quote` emojicoins.
+        base: u64,
+        /// Emojicoins per `base` octas.
+        quote: u64
+    }
```
CRBl69, line 265: Is there a reason we don't use Q64 just like for the other places ?

It would enable easy comparison.

src/move/emojicoin_arena/sources/emojicoin_arena.move
```diff
@@ -131,36 +125,169 @@ module arena::emojicoin_arena {
...
+    struct ProfitAndLoss has copy, drop, store {
+        /// Emojicoins effective value, converted to octas at current exchange rate.
+        octas_value: u128,
+        /// Unrealized gain if `octas_value` is greater than `Escrow.octas_entered`.
+        octas_gain: u128,
+        /// Unrealized loss if `octas_value` is less than `Escrow.octas_entered`.
+        octas_loss: u128,
+        /// Ratio of `octas_value` to `Escrow.octas_entered`, as a Q64.
+        octas_growth_q64: u128
+    }
```
CRBl69, line 277: Could we have `octas_initial_value` (octas value at time of enter) and `octas_current_value` (currently `octas_value`) ?

This would mean that $\\rm{octas\\_gain} = \\rm{octas\\_current\\_value} - \\rm{octas\\_initial\\_value}$, so no need to include the three other fields (`octas_{gain,loss,growth_q64}`).

I think it will also be useful to have the initial amount in the frontend.

CRBl69 (changes requested):
Look mostly good :+1:

Just a couple of DX changes.

src/move/emojicoin_arena/sources/emojicoin_arena.move
alnoki: > How would the match percentage be dynamic inside a melee

The max match percentage is the amount matched if a user joins at the beginning of the melee. For example as much as 50% match. However if they join halfway through the melee, the most they'll get is 25%

This is the linear decay model we decided on

src/move/emojicoin_arena/sources/emojicoin_arena.move
alnoki: It's possible, I just went with the ratio because this is the exact amount defined by the ratio of coins as reserves

Also makes it easier to do conversions, e.g. using a `mul_div` operation than having to cast in to Q64 etc.

If we need total order for exchange rate, thoughts on just having it get computed at indexer time by the processor?

src/move/emojicoin_arena/sources/emojicoin_arena.move
alnoki: Since users can enter multiple times, the `initial_value` descriptor isn't totally accurate, however I do understand the rationale for using a difference here and for including the reference value `octas_entered` for the frontend

Mainly the `octas_gain` field is because that is used for determining the top gains

src/move/emojicoin_arena/sources/emojicoin_arena.move
alnoki: So are you suggesting something like:

```rust
emojicoin_0_market_metadata: MarketMetadata
emojicoin_1_market_metadata: MarketMetadata
```

?

src/move/emojicoin_arena/sources/emojicoin_arena.move
alnoki: `octas_matched` renamed to `match_amount` in 8b530564

alnoki: `octas_entered` renamed to `input_amount` in 5dc785f5

src/move/emojicoin_arena/sources/emojicoin_arena.move
alnoki: Addressed per https://github.com/econia-labs/emojicoin-dot-fun/pull/478#discussion_r1901389348 https://github.com/econia-labs/emojicoin-dot-fun/pull/478#discussion_r1901390311

src/move/emojicoin_arena/sources/emojicoin_arena.move
alnoki: Addressed per https://github.com/econia-labs/emojicoin-dot-fun/pull/478#discussion_r1901389348 https://github.com/econia-labs/emojicoin-dot-fun/pull/478#discussion_r1901390311

src/move/emojicoin_arena/sources/emojicoin_arena.move
```diff
@@ -77,32 +67,34 @@ module arena::emojicoin_arena {
         max_match_percentage: u64,
         /// Maximum amount of APT to match in octas, when locking in.
         max_match_amount: u64,
-        /// Amount of rewards that are available to claim for this melee while it is still active.
-        /// Measured in octas, conditional on vault balance.
+        /// Amount of rewards that are available to claim for this melee, measured in octas, and
+        /// conditional on vault balance. Reset to 0 when cranking the schedule.
         available_rewards: u64,
         /// All entrants who have entered the melee, used as a set.
         all_entrants: SmartTable<address, Nil>,
```
alnoki, line 74: Per @davidiw

> Also a quick hint, smart table has a potential denial of service is one of the buckets gets excessively used... as you know address generation can be particularly easy to manipulate within certain constraints, so would have to check the bucketing logic to determine how easily this could result in a broken or excessively expensive smart table.
Update: since the address is hard to fully generate and it goes through another hash (siphash), the impact is likely negligible.

src/move/emojicoin_arena/sources/emojicoin_arena.move
```diff
@@ -131,36 +125,169 @@ module arena::emojicoin_arena {
...
+        /// empty `Escrow`. Inclusive of total amount matched from locking in since most recent
+        /// deposit into an empty `Escrow`. Reset to 0 upon exit.
+        input_amount: u64,
+        /// Cumulative amount of APT matched since most recent deposit into an empty `Escrow`, reset
         /// to 0 upon exit. Must be paid back in full when tapping out.
-        octas_matched: u64
+        match_amount: u64
     }
 
     struct UserMelees has key {
```
alnoki, line 141: Per @davidiw

>  it isn't clear to me why you track UserMelees on-chain when events should suffice -- as the data doesn't seem read on-chain.

alnoki: Resolved per new lightweight indexing model

src/move/emojicoin_arena/sources/emojicoin_arena.move
alnoki: Resolved per removal of `SmartTable` structs

src/move/emojicoin_arena/sources/emojicoin_arena.move
alnoki: @CRBl69 per the recent lightweight data model we discussed, this can all be done offchain. Note that I even added an example in the new README about indexing PnL and how all the information is there

alnoki: Marking resolved per eng call and new lightweight indexing

CRBl69 (approved):

PR #492, opened 2024-12-31
[ECO-2573] Denominate market cap in usd

CRBl69:
# Description

This PR denominates market cap on the home page and the market page in USD.

It uses CoinMarketCap to get the APT price in USD and falls back to showing prices in APT if the price cannot be queried.

# Testing

Go to market page and home page on vercel.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/app/dev/verify_api_keys/page.tsx
+++ b/src/typescript/frontend/src/app/dev/verify_api_keys/page.tsx
@@ -2,6 +2,7 @@ import { fetchMarketsWithCount } from "@/queries/home";
 import { AccountAddress } from "@aptos-labs/ts-sdk";
 import { VERCEL } from "@sdk/const";
 import { getAptosClient } from "@sdk/utils/aptos-client";
+import { getAptPrice } from "lib/queries/get-apt-price";
 
 export const dynamic = "force-static";
 export const revalidate = 600;
@@ -50,6 +51,13 @@ const VerifyApiKeys = async () => {
   if (res.error) {
     const msg = "\n\tLikely an invalid indexer API key.";
     throw new Error(`Couldn't fetch the price feed on the server. ${msg}`);
+  }
+
+  // Check that CoinMarketCap API key works
+  try {
+    await getAptPrice();
+  } catch (e) {
+    throw new Error(`Couldn't fetch APT price.\n\tInvalid CoinMarketCap API key.`);
   }
 
   return <div className="bg-black w-full h-full m-auto pixel-heading-2">LGTM</div>;
--- a/src/typescript/frontend/src/app/dev/verify_api_keys/page.tsx
+++ b/src/typescript/frontend/src/app/dev/verify_api_keys/page.tsx
@@ -53,11 +53,11 @@ const VerifyApiKeys = async () => {
     throw new Error(`Couldn't fetch the price feed on the server. ${msg}`);
   }
 
-  // Check that CoinMarketCap API key works
+  // Check that CoinGecko API key works
   try {
     await getAptPrice();
   } catch (e) {
-    throw new Error(`Couldn't fetch APT price.\n\tInvalid CoinMarketCap API key.`);
+    throw new Error(`Couldn't fetch APT price.\n\tInvalid CoinGecko API key.`);
   }
 
   return <div className="bg-black w-full h-full m-auto pixel-heading-2">LGTM</div>;
--- a/src/typescript/frontend/src/app/home/page.tsx
+++ b/src/typescript/frontend/src/app/home/page.tsx
@@ -14,6 +14,8 @@ import { type DatabaseModels, toPriceFeed } from "@sdk/indexer-v2/types";
 import { type DatabaseJsonType } from "@sdk/indexer-v2/types/json-types";
 import { SortMarketsBy } from "@sdk/indexer-v2/types/common";
 import { ORDER_BY } from "@sdk/queries";
+import { getAptPrice } from "lib/queries/get-apt-price";
+import { AptPriceContextProvider } from "context/AptPrice";
 
 export const revalidate = 2;
 
@@ -75,20 +77,25 @@ export default async function Home({ searchParams }: HomePageParams) {
     numMarketsPromise = getCachedNumMarketsFromAptosNode();
   }
 
-  const [priceFeedData, markets, numMarkets] = await Promise.all([
+  const aptPricePromise = getAptPrice();
+
+  const [priceFeedData, markets, numMarkets, aptPrice] = await Promise.all([
     priceFeedPromise,
     marketsPromise,
     numMarketsPromise,
+    aptPricePromise,
   ]);
 
   return (
-    <HomePageComponent
-      markets={markets}
-      numMarkets={numMarkets}
-      page={page}
-      sortBy={sortBy}
-      searchBytes={q}
-      priceFeed={priceFeedData}
-    />
+    <AptPriceContextProvider aptPrice={aptPrice}>
+      <HomePageComponent
+        markets={markets}
+        numMarkets={numMarkets}
+        page={page}
+        sortBy={sortBy}
+        searchBytes={q}
+        priceFeed={priceFeedData}
+      />
+    </AptPriceContextProvider>
   );
 }
--- a/src/typescript/frontend/src/app/market/[market]/page.tsx
+++ b/src/typescript/frontend/src/app/market/[market]/page.tsx
@@ -6,6 +6,8 @@ import { pathToEmojiNames } from "utils/pathname-helpers";
 import { fetchChatEvents, fetchMarketState, fetchSwapEvents } from "@/queries/market";
 import { getMarketAddress } from "@sdk/emojicoin_dot_fun";
 import { type Metadata } from "next";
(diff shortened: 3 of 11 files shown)


src/typescript/frontend/src/app/dev/verify_api_keys/page.tsx
```diff
@@ -52,6 +53,13 @@ const VerifyApiKeys = async () => {
     throw new Error(`Couldn't fetch the price feed on the server. ${msg}`);
   }
 
+  // Check that CoinMarketCap API key works
```
alnoki, line 56: - [ ] Try CoinGecko, ideally without an API key if possible
- [ ] Mark as resolved this comment if able to use it

src/typescript/frontend/src/app/dev/verify_api_keys/page.tsx
```diff
@@ -52,6 +53,13 @@ const VerifyApiKeys = async () => {
     throw new Error(`Couldn't fetch the price feed on the server. ${msg}`);
   }
 
+  // Check that CoinGecko API key works
+  try {
+    await getAptPrice();
+  } catch (e) {
+    throw new Error(`Couldn't fetch APT price.\n\tInvalid CoinGecko API key.`);
```
alnoki, line 60: - [ ] If unable to fetch price at runtime (either because key is invalid or not present), fall back to APT market cap behavior
- [ ] Fail at build time if key is provided but invalid
- [ ] Allow build if API key not provided
- [ ] Resolve this comment once all above addressed

CRBl69: Updated behaviour:

- API key is only checked at build time on Vercel.
- At runtime, if the key does not work, falls back to showing prices in APT.

alnoki: @CRBl69 I believe these updates address the above items, but yielding to you to check them off. I'll leave an approving review so once you check them all off and mark this as resolved you can merge

alnoki (approved):

alnoki (approved):

src/typescript/frontend/src/lib/queries/get-apt-price.ts
```diff
@@ -0,0 +1,31 @@
+"use server";
+
+import { COINGECKO_API_KEY } from "lib/server-env";
```
alnoki, line 3: - [ ] Document this in example.env as I said here: https://github.com/econia-labs/emojicoin-dot-fun/pull/492#discussion_r1901981740

src/typescript/frontend/src/app/home/page.tsx
```diff
@@ -75,20 +77,25 @@ export default async function Home({ searchParams }: HomePageParams) {
...
   return (
-    <HomePageComponent
-      markets={markets}
-      numMarkets={numMarkets}
-      page={page}
-      sortBy={sortBy}
-      searchBytes={q}
-      priceFeed={priceFeedData}
-    />
+    <AptPriceContextProvider aptPrice={aptPrice}>
```
xbtmatt, line 90: This should definitely go in `providers.tsx`, otherwise it's conditional based on whether or not the user is visiting the home page, but also because `providers.tsx` wraps the global `layout.tsx` or vice versa (can't remember)

src/typescript/frontend/src/app/market/[market]/page.tsx
```diff
@@ -6,6 +6,8 @@ import { pathToEmojiNames } from "utils/pathname-helpers";
 import { fetchChatEvents, fetchMarketState, fetchSwapEvents } from "@/queries/market";
 import { getMarketAddress } from "@sdk/emojicoin_dot_fun";
 import { type Metadata } from "next";
+import { getAptPrice } from "lib/queries/get-apt-price";
+import { AptPriceContextProvider } from "context/AptPrice";
```
Comment (xbtmatt) [inline src/typescript/frontend/src/app/market/[market]/page.tsx:10]:
```suggestion
```

src/typescript/frontend/src/app/market/[market]/page.tsx
```diff
@@ -75,23 +77,26 @@ const EmojicoinPage = async (params: EmojicoinPageProps) => {
...
-        data={{
-          symbol: state.market.symbolData.symbol,
-          swaps,
-          chats,
-          state,
-          marketView,
-          ...state.market,
-        }}
-      />
+      <AptPriceContextProvider aptPrice={aptPrice}>
```
Comment (xbtmatt) [inline src/typescript/frontend/src/app/market/[market]/page.tsx:88]:
```suggestion
```

src/typescript/frontend/src/app/market/[market]/page.tsx
```diff
@@ -75,23 +77,26 @@ const EmojicoinPage = async (params: EmojicoinPageProps) => {
...
+          data={{
+            symbol: state.market.symbolData.symbol,
+            swaps,
+            chats,
+            state,
+            marketView,
+            ...state.market,
+          }}
+        />
+      </AptPriceContextProvider>
```
Comment (xbtmatt) [inline src/typescript/frontend/src/app/market/[market]/page.tsx:99]:
```suggestion
```

src/typescript/frontend/src/components/pages/emojicoin/components/main-info/MainInfo.tsx
```diff
@@ -96,6 +98,10 @@ const MainInfo = ({ data }: MainInfoProps) => {
     }
   }, [stateEvents]);
 
+  const aptPrice = useAptPrice();
+
+  const usdMarketCap = aptPrice ? toNominal(marketCap) * aptPrice : undefined;
```
xbtmatt, line 103: This should be in its own hook, that way you can deduplicate the logic, since it's copy pasted a few times

src/typescript/frontend/src/components/pages/home/components/main-card/MainCard.tsx
```diff
@@ -60,6 +62,10 @@ const MainCard = (props: MainCardProps) => {
     };
   }, [featured]);
 
+  const aptPrice = useAptPrice();
+
+  const usdMarketCap = aptPrice ? toNominal(marketCap) * aptPrice : undefined;
```
xbtmatt, line 67: swap with `useMarketCap` hook (mentioned above)

src/typescript/frontend/src/components/pages/home/components/table-card/TableCard.tsx
```diff
@@ -81,6 +83,10 @@ const TableCard = ({
     };
   }, [sortBy, animationEvent, staticVolume24H, staticMarketCap]);
 
+  const aptPrice = useAptPrice();
+
+  const usdMarketCap = aptPrice ? toNominal(marketCap) * aptPrice : undefined;
```
xbtmatt, line 88: swap with `useMarketCap` hook (mentioned above)

src/typescript/frontend/src/lib/queries/get-apt-price.ts
```diff
@@ -0,0 +1,31 @@
...
+      }
+    )
+      .then((r) => r.json())
+      .then((r) => r.aptos.usd as number)
+      .catch((e) => {
+        console.error("Could not get APT price from CoinGecko.", e);
+        return undefined;
+      }),
+  ["apt-price"],
+  { revalidate: 60 * 10 } // Ten minutes
```
xbtmatt, line 30: What's the API rate limit? I think this could/should just be 1 minute

xbtmatt (changes requested):
1. Move context to the `providers.tsx` file, noted in one of the comments
2. Move market cap logic to `useMarketCap` hook and re-use it to avoid duplicating that logic
3. Make sure everything works without an API key and that all values default to non-USD prices when no API key is provided

Looks good for the most part though :)

xbtmatt (commented):
Two things left after reviewing on eng call:

1. Move market cap logic to a hook
2. Use the new rate limit info to change the polling interval (if the rate limit is generous enough to withstand 1 min per call)

src/typescript/frontend/src/hooks/use-usd-market-cap.ts
```diff
@@ -0,0 +1,9 @@
+import { useAptPrice } from "context/AptPrice";
+import { toNominal } from "lib/utils/decimals";
+
+// Returns market cap in USD.
+// If the APT price isn't available, undefined will be returned.
```
xbtmatt, line 5:
```suggestion
/**
 * Returns the market cap in USD.
 *
 * If the APT price isn't available, it will return `undefined`.
 */
```

src/typescript/frontend/src/lib/queries/get-apt-price.ts
xbtmatt, line 30:
```suggestion
  { revalidate: 10 } // Ten seconds.
```

src/typescript/frontend/src/app/home/page.tsx
xbtmatt: Determined on call that we'll keep this the same- it's too difficult to change. The reason it's done this way is because we need to pass server data to the layout, and the layout providers.tsx is technically a client component. There's no easy way to get around this other than a global zustand store, which isn't worth the effort + re-rendering. It makes more sense to just wrap the components that need it with the provider.

xbtmatt (approved):

PR #497, opened 2025-01-06
[ECO-2650] Fix size of scramble to avoid elements moving

CRBl69:
# Description

This PR sets the width of button to a fix size, to avoid the buttons changing size when the scrable animation happens. This prevents elements (especially in flexes) to shift around.

# Testing

Check all buttons hover animation.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/components/button/index.tsx
+++ b/src/typescript/frontend/src/components/button/index.tsx
@@ -68,27 +68,53 @@ const Button = <E extends React.ElementType = "button">({
             })}
 
           {!isScramble ? (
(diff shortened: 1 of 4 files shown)


CRBl69:
- [ ] fix connect button in the mobile hamburger menu

xbtmatt (approved):

xbtmatt (changes requested):
Last thing to fix here is the spacing on mobile, there's a character's worth of space here that we can/should get rid of on the left of the mobile / hamburger menu page for the connect wallet button

xbtmatt (approved):

PR #498, opened 2025-01-06
[ECO-2631] Add coingecko routes

CRBl69:
# Description

This PR adds the routes CoinGecko needs to list emojicoin.fun

# Testing

Query both routes on the `/coingecko` parent route.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/app/coingecko/tickers/route.ts
+++ b/src/typescript/frontend/src/app/coingecko/tickers/route.ts
@@ -1,10 +1,34 @@
+import { APTOS_COIN_TYPE_TAG } from "@sdk/const";
 import { postgrest } from "@sdk/indexer-v2/queries/client";
 import { TableName } from "@sdk/indexer-v2/types";
-import Big from "big.js";
+import { toNominalPrice } from "@sdk/utils";
+import { toNominal } from "lib/utils/decimals";
 import type { NextRequest } from "next/server";
 import { stringifyJSON } from "utils";
+import { estimateLiquidityInUSD } from "./utils";
+
+// Since this is a public endpoint, set revalidate to 1 to ensure it can't spam the indexer.
+export const revalidate = 1;
+
 /**
+ * Query params:
+ *  - skip: number, the number of rows to skip at the beginning, default is 0
+ *  - limit: number, the max number of rows per query, default is 100, max is 500
+ *
  * @see {@link https://docs.google.com/document/d/1v27QFoQq1SKT3Priq3aqPgB70Xd_PnDzbOCiuoCyixw/edit?tab=t.0}
+ *
+ * @example
+ * // The example for a single row, given in the Google doc linked above:
+ * ```json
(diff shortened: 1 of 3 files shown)


xbtmatt (commented):
Can you link in the PR description the API specs for this? And maybe in a jsdoc comment on route.ts on both endpoints or in a README.md

CRBl69:
https://docs.google.com/document/d/1v27QFoQq1SKT3Priq3aqPgB70Xd_PnDzbOCiuoCyixw/edit?tab=t.0

xbtmatt (changes requested):
I think we still need the circulating/total supply endpoint, right?

The `/tickers` endpoint is different for an AMM DEX:

```json
"ticker_id": "0x55d398326f99059ff775485246999027b3197955_0xbb4cdb9cbd36b01bd1cbaebf2de08d9173bc095c",
"base_currency": "0x55d398326f99059ff775485246999027b3197955",
"target_currency": "0xbb4cdb9cbd36b01bd1cbaebf2de08d9173bc095c",
“pool_id”:
“0x36696169c63e42cd08ce11f5deebbcebae652050”,
"last_price":"50.0",
"base_volume":"10",
"target_volume":"500",
“liquidity_in_usd”:“100”,
"bid":"49.9",
"ask":"50.1",
```

It also says we have to provide the +2/-2% depth but that's actually really complicated so we can just wait for them to ask for it if it's really necessary.

We do, however, need the `liquidity_in_usd` field, for which I think you can just use the virtual reserves if it's in the bonding curve and the real reserves if it's post bonding curve, convert base to quote, add to quote, then multiply the sum of that by the Aptos USD price

xbtmatt (approved):
I added the changes, should be good now, I'll make sure it's correct before merging.

src/typescript/frontend/src/app/coingecko/tickers/route.ts
```diff
@@ -0,0 +1,67 @@
...
+  const APTOS_COIN = APTOS_COIN_TYPE_TAG.toString();
+
+  const data = markets.data?.map((e) => ({
+    ticker_id: `${e.market_address}::coin_factory::Emojicoin_${APTOS_COIN}`,
+    base_currency: `${e.market_address}::coin_factory::Emojicoin`,
+    target_currency: APTOS_COIN,
+    pool_id: e.symbol_emojis.join(""),
+    last_price: toNominalPrice(e.last_swap_avg_execution_price_q64).toString(),
+    base_volume: toNominal(BigInt(e.daily_volume)).toString(),
+    target_volume: toNominal(BigInt(e.daily_base_volume)).toString(),
```
xbtmatt, line 62: Is this correct? I just realized we're using daily (quote) volume for `base_volume` and base volume for `target_volume`.

CRBl69: Addressed in 708b9b8.

xbtmatt (approved):

PR #508, opened 2025-01-16
[ECO-2664] Add home page arena update

CRBl69:
# Description

This PR updates the home page for the new arena feature.

# Testing

See home page.

# Checklist

### DIFF
--- a/cfg/cspell-frontend-dictionary.txt
+++ b/cfg/cspell-frontend-dictionary.txt
@@ -56,3 +56,4 @@ bytea
 nominalize
 dexscreener
 clippyts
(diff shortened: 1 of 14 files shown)


xbtmatt (commented):
Consider gating this entire PR by env var "ARENA_ENABLED", just a note for self

xbtmatt (commented):
One last thing to fix: fixing the black background behind emojis when it loads

xbtmatt (approved):
Only thing to note is that the fetchMarketStateByAddress calls can be combined, but I'll do it in another PR!

PR #509, opened 2025-01-16
[ECO-2666] Update arena README publish, PnL sections

alnoki:
# Description

Update emojicoin arena README:

1. Remove PnL pseudo code now that it is covered in ongoing indexing
1. Add section with publish commands, filing https://github.com/aptos-labs/aptos-core/issues/15769

# Testing

1. Published on testnet: https://explorer.aptoslabs.com/txn/6596176228?network=testnet

### DIFF
--- a/src/move/emojicoin_arena/README.md
+++ b/src/move/emojicoin_arena/README.md
@@ -65,18 +65,6 @@ writeset-aware offchain processing. In order words, the data model is designed
 to enable full indexing using only writesets and events, with as few struct and
 event fields as possible.
 
-As an example, calculating PnL offchain for a given escrow looks something like:
-
-```python
-profit = sum_over_all_events_since_escrow_was_last_empty(
-    Enter.input_amount + Enter.match_amount
-) / current_value_of_holdings()
-```
-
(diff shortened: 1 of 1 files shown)


src/move/emojicoin_arena/README.md
```diff
@@ -65,24 +65,43 @@ writeset-aware offchain processing. In order words, the data model is designed
...
+aptos move publish \
+    --assume-yes \
+    --max-gas 500000 \
+    --named-addresses $NAMED_ADDRESSES \
+    --profile $PROFILE
+```
+
+Note that the `--max-gas` argument is required to prevent issues with the
+simulator reporting that the `emojicoin_arena` module doesn't exist (even though
+it shouldn't exist, because it's the one being published).
```
xbtmatt, line 104: This is the `Failed to convert transaction data from storage` error, right? Or is it a different error code

Could you put the error message in the comment here just in case I ever `ctrl + f` for it again?

xbtmatt (approved):

src/move/emojicoin_arena/README.md
alnoki: Addressed in 3300ca5a

PR #522, opened 2025-01-25
[ECO-2700] Clean up environment variables, fix documentation, remove unused `frontend` container in Docker

xbtmatt:
# Description

The Docker container for the `frontend` service had a ton of repeated variables that made the environment variable files look huge, so I removed it because we never use it (we disabled the frontend tests).

I left the frontend testing code in there in case we ever do want to use it again- just removed the option to run a frontend container in Docker, since it was unnecessary and made the code/configuration for Docker needlessly confusing.

I updated all the env files to have a corresponding network and sample values, so that it's easy to run very quickly.

# Testing

Run through the examples manually to see if it actually works.

For `local` network:

```shell
# Navigate to `src/docker` and run:
pnpm run docker:prune
cd src/docker
docker compose -f compose.local.yaml --env-file example.local.env up -d

# Run the frontend for localnet
cd src/typescript
mv example.local.env .env
pnpm run dev
```
- [x] Test on localnet

`testnet`:
```shell
# For testnet:
pnpm run docker:prune
cd src/docker
docker compose -f compose.yaml --env-file example.testnet.env up -d

# Run the frontend for testnet
cd src/typescript
cp example.testnet.env .env
pnpm run dev
```

- [x] Test it for testnet
- [x] Test it for mainnet as well

And it should work with the locally running testnet emojicoin indexer.

# Checklist


src/docker/README.md
```diff
@@ -9,20 +9,26 @@ build the processor, otherwise you will get errors.
 git submodule update --init --recursive
 ```
 
-Ensure that your environment variables are set, typically with an `.env`
```
alnoki, line 12: - [ ] Merge blocking dependency cascade beginning with ECO-2695 before this merges, so that we are not sharing out instructions for a stack that will ultimately fail if run against mainnet (because the current processor in the hanging release needs to get resolved)

src/docker/README.md
```diff
@@ -9,20 +9,26 @@ build the processor, otherwise you will get errors.
...
-
-Then you can follow the simple examples below showing how to run the processor
-with and without an Aptos local fullnode.
-
-## Run the indexer processor by itself
+## Run the indexer processor by itself for `mainnet` or `testnet`
 
 ```shell
-docker compose -f compose.yaml up
+docker compose -f compose.yaml up --env-file example.testnet.env up -d
```
CRBl69, line 15: Does this work ? `docker ... up ... up ...`. You have two `up`s in there.

src/docker/example.testnet.env
```diff
@@ -1,55 +1,43 @@
...
 # devnet: https://grpc.devnet.aptoslabs.com:443
-# local: http://host.docker.internal:50051
+# local: http://host.docker.internal:50051 (if running a localnet)
 #
-# NOTE: If you get errors connecting to `host.docker.internal` on Docker Desktop
-# for WSL 2, see the explanation in `example.local.env` on which configuration
-# settings you need to enable.
+# ---------------------------------------------------------------------------- #
 GRPC_DATA_SERVICE_URL="https://grpc.testnet.aptoslabs.com:443"
+GRPC_AUTH_TOKEN="MUST_PROVIDE_TOKEN_HERE"
```
CRBl69, line 36: This is minor but could we have consistent casing/formatting for when we put placeholders in env vars ? I see you use both `this_formatting` and `THIS_FORMATTING`.

CRBl69 (changes requested):
Minor stuff to fix.

~Also, I don't personally use it but is there a reason to remove the frontend from docker compose ?~ Never mind, just read the PR description.

src/docker/README.md
xbtmatt:
```suggestion
docker compose -f compose.yaml --env-file example.testnet.env up -d
```

src/docker/example.testnet.env
xbtmatt: Addressed in fb83859

CRBl69 (approved):

PR #533, opened 2025-01-30
[ECO-2725] Update `localnet` and `deployer` to use Aptos CLI `v6.0.1`

xbtmatt:
# Description

The CLI version needs to bump now that AMD/ARM successfully built the CLI v6.0.1.

And the arena code needs to deploy for local testing because it uses `--move-2`.

### DIFF
--- a/.github/workflows/push-deployer.yaml
+++ b/.github/workflows/push-deployer.yaml
@@ -32,7 +32,7 @@ jobs:
         context: '.'
         file: 'src/docker/deployer/Dockerfile'
         labels: '${{ steps.metadata.outputs.labels }}'
-        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
+        platforms: 'linux/amd64'
         push: 'true'
         tags: '${{ steps.metadata.outputs.tags }}'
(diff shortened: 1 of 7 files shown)


src/docker/deployer/sh/init-profile.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+# cspell:word obase
```
xbtmatt, line 2:
```suggestion
```

.github/workflows/push-deployer.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
         context: '.'
         file: 'src/docker/deployer/Dockerfile'
         labels: '${{ steps.metadata.outputs.labels }}'
-        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
+        platforms: 'linux/amd64'
```
xbtmatt, line 35:
```suggestion
        platforms: '${{ vars.DOCKER_IMAGE_PLATFORMS }}'
```

CRBl69 (approved):

PR #538, opened 2025-01-31
[ECO-2721] Fix SDK timeout error

CRBl69:
# Description

The SDK tests often fail due to a timeout.

This PR fixes that

# Testing

See SDK tests pass 100% of the time

# Checklist


xbtmatt:
Looking over the SDK test actions for the past 3-4 days, it looks like this issue has finally been resolved with the recent fix! If it starts happening again we can open another PR but this seems to be resolved now.

PR #542, opened 2025-02-03
[ECO-2735] Update GitHub action to parallelize multi-platform build on ARM runner

alnoki:
# Description

- [x] Per https://github.com/econia-labs/cloud-infra/pull/20, update the `push-broker.yaml` action to parallelize across the new GitHub actions ARM runners
- [x] Bump `rust-builder` version number

# Testing

- [ ] Successfully push new tags once https://github.com/econia-labs/aptos-indexer-processors/pull/82 merges, with bumped submodule


.github/workflows/push-broker.yaml
```diff
@@ -2,64 +2,82 @@
...
+        REPO_NAME="emojicoin-dot-fun-indexer-broker"
+        BASE_URL="https://hub.docker.com/v2/repositories/\
+        ${DOCKER_ORG}/\
+        ${REPO_NAME}/\
+        tags"
+        BASE_URL=$(echo "$BASE_URL" | tr -d ' ')
+        echo "url=$BASE_URL" >> "$GITHUB_OUTPUT"
+    - name: 'Append ARM images to AMD manifest, remove old ARM images'
+      # Loop over all the AMD64 tags and append the ARM64 tags to the manifest,
+      # then delete the old ARM64 tags from Docker Hub.
```
xbtmatt, line 69: How come you loop over all the AMD64 tags? Why not just for this image? Or am I misunderstanding

.github/workflows/push-broker.yaml
```diff
@@ -2,64 +2,82 @@
...
       run: |
-        echo "${{ steps.metadata.outputs.tags }}" | while read -r tag; do
+        echo "${{ needs.build.outputs.amd64-tags }}" | while read -r tag; do
         if [ ! -z "$tag" ]; then
-        docker buildx imagetools create --append -t "$tag" "${tag}-arm64"
+        docker buildx imagetools create --append -t "$tag" "$tag-arm64"
+        VERSION=$(echo "$tag" | cut -d':' -f2)
+        curl -X DELETE \
+        -H "Authorization: JWT ${{ secrets.DOCKERHUB_TOKEN }}" \
+        "${{ steps.base-url.outputs.url }}/${VERSION}-arm64/"
```
xbtmatt, line 77: I think the safest solution here to pass the secret here is [through an environment variable](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#example-using-bash), otherwise `curl` + actions logs could accidentally leak the token

CRBl69: I think GitHub automatically censors secrets in output logs.

alnoki: Actually now that I'm looking through the [docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#example-using-bash), it looks like the recommendation of an environment variable was mainly just to avoid quoting issues:

> If you must pass secrets within a command line, then enclose them within the proper quoting rules. Secrets often contain special characters that may unintentionally affect your shell. To escape these special characters, use quoting with your environment variables. For example:

.github/workflows/push-broker.yaml
alnoki: @xbtmatt there are multiple generated Docker image tags: the one from the git tag, e.g. `6.0.2`, and `latest`

This merges images from `6.0.2-arm64` *and* `latest-arm64`

.github/workflows/push-broker.yaml
alnoki: Per @CRBl69 , consider using `curl -H @headers.txt https://example.com`

alnoki: @xbtmatt addressed as of beef903d

xbtmatt (approved):

PR #496, opened 2025-01-06
[ECO-2637] Add blank candles on missing data

CRBl69:
# Description

Add blank candles where no candles are present.

# Testing

See market page with inactivity.

# Checklist

- [ ] Did you check all checkboxes from the linked Linear task?

Issue #486, opened 2024-12-28
tradingview missing candles 


MoonShiesty:
this token launched nov 20, but there aren't 40 1d candles. looks like empty candles aren't being created by the TV adapter

https://www.emojicoin.fun/market/1st_place_medal

![image](https://github.com/user-attachments/assets/4006262b-8b62-4084-8639-5c28e3daef6e)

alnoki:
@MoonShiesty thank you for filing an issue! Noting that we are slating this into our current queue of aesthetic fixes (internally it is tagged ECO-2637)

CRBl69:
It seems that this is default TradingView (and other charts) behaviour.

The screenshot below is the chart of the [DJI](https://finance.yahoo.com/chart/%5EDJI) (Dow Jones Industrial) during September 2001. After 9/11, the market was closed for a couple of days (11 to 16). The chart shows as continuous, skipping the closed days.

![image](https://github.com/user-attachments/assets/2fb9f791-3ba5-4d28-a1b7-6007976db5d5)

The same behavior can be observed in the screenshot you provided. During days with 0 activity, no bars were generated. When showing the chart, all existing bars are put one next to the other, without leaving blank spaces.

It seems though that skipping empty activity periods is possible (see [this comment on the trading view repo](https://github.com/tradingview/lightweight-charts/issues/1288#issuecomment-1453254487)).

@alnoki what are your thoughts on this ? Should we add a space when there are missing candles or not ?

MoonShiesty:
this comparison doesn't make sense, emojicoin's market never closes

ive traded on many platforms, this is the first to omit empty candles

the tradingview charting library even has trouble with default window sizing because it expects more candles than are drawn in the window

MoonShiesty:
> It seems though that skipping empty activity periods is possible (see this comment on the trading view repo).


right now you ARE skipping empty activity periods. instead, the candle should have zero volume, which correctly represents what occurs in the market data (the market was no closed). or course this is possible, every other TV integration operates this way

MoonShiesty:
here is an example of empty candles from hyperliquid

the more i think about this the less sense it makes, for example, if i switch to the 1s candle, the entire chart scaling will be nonsense since 99.999% of the candles will be missing

<img width="1198" alt="image" src="https://github.com/user-attachments/assets/861a966b-4ae6-4e9d-b314-98003febb928" />

alnoki:
[...]

PR #569, opened 2025-02-07
fix(ECO-2637): Address missing candles issue

0xForkh:
This is a contribution to PR #496 which addresses issue #486 
 
After the changes from PR #496, candles are displayed for every minute, even when there is no trade, which results to most graphs being full of empty candles in the default view (1m). For example, if there are no transaction in a 1h timeframe, there will be 60 empty candles on the graph.

I checked how it works on other platforms, and usually the default view is 1h.

So this PR changes the default view to 1h instead of 1m.


xbtmatt (commented):
The main thing I wanted to check with this is that you get no errors with `has_empty_bars: true`, because for some reason a while ago I used to get errors with that. It might have something to do with the way we process the candlesticks.

You might need to test this on `testnet` with a new market you create with some empty candlesticks and whatnot just to make sure there aren't any issues

xbtmatt (commented):
Nice, looks like it's working- but for some of the charts there are so many empty candles that it's difficult to see the actual price action, which is exactly why I didn't implement it originally 😅 

Can you try adding a simple toggle for this?

I can come up with a design freehand once you're done (or you can if you want), just a simple toggle button, like we have for the "Animate" stuff on the home page.

But then we can store their setting in `localStorage` and use that as their preference

With that I think this would be perfect!

src/typescript/frontend/src/components/charts/PrivateChart.tsx
```diff
@@ -111,9 +116,27 @@ export const Chart = (props: ChartContainerProps) => {
         onResultReadyCallback(symbols);
       },
       resolveSymbol: async (symbolName, onSymbolResolvedCallback, _onErrorCallback) => {
+        // We pass parameters in the symbolName to prevent re-rendering.
+        // This seems to be the only way to seamlessly change some properties such as has_empty_bars without re-creating the whole chart.
+        // Normally, we would add the state showEmptyCandles in the dependency array to re-create the whole datafeed and the chart.
+        // But this causes the re-creation of the whole chart, and briefly breaks the view (for ex: the custom button gets removed and re-added).
+        // This solution allows to call chart.setSymbol with custom params to update the chart seamlessly.
+        // Uses similar syntax to url query params.
```
xbtmatt, line 124:
```suggestion
        // We pass parameters in the symbolName to prevent re-rendering.
        // This seems to be the only way to seamlessly change some properties such as has_empty_bars without re-creating the whole chart.
        // Normally, we would add the state showEmptyCandles in the dependency array to re-create the whole datafeed and the chart.
        // But this causes the re-creation of the whole chart, and briefly breaks the view (for ex: the custom button gets removed and re-added).
        // This solution allows to call chart.setSymbol with custom params to update the chart seamlessly.
        // Uses similar syntax to url query params.
```

When you put this into the other function, at the top of the jsdoc comment for it, just state what it does (passes data into the symbol info, and we remove it later) and then explain why below that

xbtmatt (commented):
Looks great :) Like we discussed on call,

1. Move the logic for parsing the new symbol info structure into a function outside of the component
2. Move the logic for creating the new symbol info outside the component as well

Just to make it easier to refactor and use it in other places

src/typescript/frontend/src/hooks/use-state-persisted.ts
```diff
@@ -0,0 +1,49 @@
+import { useEffect, useState } from "react";
+
+// hook "useLocalStorage" from package "react-use" doesn't seem to provide the correct previous value when you set a state by using a callback function.
+// This custom hook behaves exactly like useState from React, but it persists the state in the local/session storage.
```
xbtmatt, line 4: Just note here that this should be only used in a single place or with a completely unique key, because it doesn't handle setting the state in multiple places

src/typescript/frontend/src/hooks/use-state-persisted.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @template T The type of the state value
+ * @param {string} key The key under which the value will be stored in storage
+ * @param {T} initialValue The initial value to use if no value exists in storage
+ * @param {"localStorage" | "sessionStorage"} [storageType="localStorage"] The type of storage to use
+ * @returns {[T, React.Dispatch<React.SetStateAction<T>>, () => void]} A tuple containing:
+ *   - The current state value
+ *   - A function to update the state
+ *   - A function to clear the stored value
+ */
+function useStatePersisted<T>(
```
xbtmatt, line 20: I feel *really* bad because this is awesome documentation- really, it's incredible. But I should have pointed you at our existing `userSettingsStore` earlier, which is actually fairly straightforward to use and uses semver with `localStorage` to invalidate stale caches with old types.

Honestly, I forgot how robust the solution @CRBl69 made was, and it really makes more sense to just use that, since `zustand` will handle the state globally as well.

https://github.com/econia-labs/emojicoin-dot-fun/blob/8e342c80230d83c72bcd9882ed9672c21c0201ed/src/typescript/frontend/src/lib/store/user-settings-store.ts#L11

Could you refactor this to use that? You would just need to add a `showEmptyBars` and a `setShowEmptyBars` in that provider, and then bump the `settings` version to `2.0.0` [here](https://github.com/econia-labs/emojicoin-dot-fun/blob/f09c3a838de99b6a8a41dedd30398cdabb3f1fe7/src/typescript/frontend/src/configs/local-storage-keys.ts#L13)

xbtmatt (commented):
Just wanna say this really is awesome- it's really nice to have a framework for adding config options to the chart now. I would love to get it in, but I do think we should refactor this to use hte existing local storage solution. I commented below, but I apologize again for not suggesting it earlier!

Once that's updated this is good to go 🎉

PR #574, opened 2025-02-10
[ECO-2712] Fix/update stats page

xbtmatt:
# Description

Updated the stats page to be rendered server-side entirely, with proper pagination and styling.

There are a lot of new re-usable components here that we can use elsewhere, noting here to refactor/use them later:

- Table styling for a trade history page
- Server side emojis with proper font class names

Checklist, some may be moved to a follow-up task:
- [ ] Re-style per @kiki 
- [X] (Passing too much data from server to client, so page can't be cached)
  - [X] Maybe paginate it?
- [ ] Update dropdown for per market/global to be like the wallet connection one
- [ ] Put a header on the homepage between home and pools, handle resizing accordingly
- [ ] Also add in a field that allows users to get the cumulative integrator revenue by integrator address
  - [ ] For example, they could input an integrator address and see how much revenue it has had from swaps and from market registrations

# Testing

A good amount of manual testing and personal use (it's fun to look at!)

# Checklist


xbtmatt:
Removing stale flag

PR #579, opened 2025-02-11
[ECO-2807] Have the `deployer` service start a very short arena; add tests to set next melee duration

xbtmatt:
# Description

- [X] Add a script in the build publish payload command in the `deployer` service to find the default duration and replace it with a single microsecond
- [X] Ensure that the arena immediately ends in tests
- [X] Add helper functions to end the very first melee and start a new one, with all unlocked (no grace period) markets
- [x] Re-build and push the new `deployer` image with this PR

# Testing

- [x] Ensure the general test flow and emojicoin client continue to work
- [x] Ensure websockets tests continue to work
- [x] Ensure the next duration of the arena is properly set and used

# Checklist


src/typescript/sdk/src/const.ts
```diff
@@ -119,6 +119,16 @@ export const MARKET_REGISTRATION_DEPOSIT = 1n * ONE_APT_BIGINT;
 export const MARKET_REGISTRATION_GAS_ESTIMATION_NOT_FIRST = ONE_APT * 0.005;
 export const MARKET_REGISTRATION_GAS_ESTIMATION_FIRST = ONE_APT * 0.6;
 export const BASIS_POINTS_PER_UNIT = 10_000n;
+
+// Arena constants.
+export const ARENA_CONSTANTS = {
+  // Note that in test, the `deployer` service changes this value to `1n`; i.e., one microsecond.
```
alnoki, line 125: - [ ] Update to something more generalized such that this comment does not need to be revised in case the deployer service changes
- [ ] Resolve this comment

src/typescript/sdk/tests/e2e/broker/utils.ts
```diff
@@ -126,19 +133,73 @@ export const customWaitFor = async (condition: () => boolean) =>
   });
 
 /**
- * Have the publisher trade on both markets it registered to unlock them by ending the grace period.
+ * Have the publisher register and trade on a market to unlock it by ending the grace period.
+ *
+ * This facilitates beginning a new arena, since the new arena must have a new unique combination
+ * of market IDs.
+ *
+ * This function can/should only be called *ONCE* during the entire jest test suite, since it makes
```
alnoki, line 141: Is there any way to enforce this?

src/typescript/sdk/tests/e2e/broker/utils.ts
```diff
@@ -126,19 +133,73 @@ export const customWaitFor = async (condition: () => boolean) =>
...
+ * assumptions about the initial state on-chain.
+ */
+export const registerAndUnlockMarketForTestSuite = async (newSymbol: SymbolEmoji[]) => {
+  const emojicoin = new EmojicoinClient();
+  const publisher = getPublisher();
+  const numMarkets = await fetchNumRegisteredMarkets();
+  // This function should only be called at the very beginning of the test suite initialization,
```
alnoki, line 148: Is there any way to enforce this?

src/typescript/sdk/tests/e2e/broker/utils.ts
xbtmatt: It will be enforced because it will fail if it's ever called twice- the documentation is there to explain why it errors out

src/typescript/sdk/src/const.ts
xbtmatt: I was just looking at this comment myself and wondering if this might ever change, haha, will update it.

xbtmatt: Fixed 8a9d136288f1a6e80302cfe04c2c04d7193340df

src/typescript/sdk/tests/e2e/broker/utils.ts
xbtmatt: Okay unfortunately there's no easy way to do it- I just made it very clear what it's doing and how to use it, should be good enough

xbtmatt: I ended up separating all the tests into three different iterations- so they can't interfere with each other anymore. Discussed on call and resolved

PR #581, opened 2025-02-11
test(ECO-2808): Add E2E tests for arena processor refactor

CRBl69:
# Description

This PR adds E2E tests for the arena processor refactor.

# Testing

Run arena E2E tests.

# Checklist


xbtmatt (approved):
Trying to clear these out today, just going to note two changes I made:

1. I moved the utilities into an `arena/utils.ts` file
2. I removed the wait for melee ending from the `crankAndSetNextMeleeDuration` function- since it was doing too much. You now have to explicitly wait for the melee to end before calling it
3. Converted title/commits to conventional commits so it's mergeable

PR #593, opened 2025-02-13
[ECO-2828] Add shadcn styles to `global.css` for prototyping

xbtmatt:
# Description

Adds shadcn styles to `global.css` and tailwind utilities so we can use shadcn components while prototyping and in test

### DIFF
--- a/src/typescript/frontend/src/app/global.css
+++ b/src/typescript/frontend/src/app/global.css
@@ -26,7 +26,7 @@
   --warning: #ffb119;
   --error: #f3263e;
 
-  /* ShadCN styles- currently only used for testing and prototyping. */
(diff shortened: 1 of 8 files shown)


src/typescript/frontend/src/app/global.css
```diff
@@ -25,6 +25,28 @@
   --pink: #cd2f8d;
   --warning: #ffb119;
   --error: #f3263e;
+
+  /* ShadCN styles- currently only used for testing and prototyping. */
```
xbtmatt, line 29:
```suggestion
  /* shadcn styles- currently only used for testing and prototyping. */
```

CRBl69 (approved):

PR #594, opened 2025-02-13
[ECO-2817] Pin the broker and processor versions in `compose.yaml`

xbtmatt:
- [x] Pin versions in `compose.yaml`
- [x] Remove fields added for `arena` that aren't in `4.0.2`


src/typescript/sdk/src/indexer-v2/types/postgres-numeric-types.ts
```diff
@@ -104,11 +104,9 @@ export const floatColumns: Set<AnyColumnName> = new Set([
   "emojicoin_0_proceeds",
   "emojicoin_0_exchange_rate_base",
   "emojicoin_0_exchange_rate_quote",
-  "emojicoin_0_market_id",
```
alnoki, line 107: What is this about? Just didn't see it in PR description

src/typescript/sdk/tests/e2e/schema.test.ts
```diff
@@ -60,6 +60,8 @@ describe("verifies the schema is what's expected", () => {
     const floats: Set<AnyColumnName> = new Set();
     const integers: Set<AnyColumnName> = new Set();
     for (const table of tables) {
+      expect(definitions[table]).toBeDefined();
```
alnoki, line 63: Ditto for this

alnoki (approved):

src/typescript/sdk/tests/e2e/schema.test.ts
xbtmatt: When the schema test fails it's not clear why because `jest` error logs are bad when the error isn't `expect(...)`ed, so I tried to make it more clear.

PR #596, opened 2025-02-13
[ECO-2820] Bump frontend/SDK versions

xbtmatt:
# Description

- [x] Since this only adds to community projects JSON data and fixes the market cap representation, it is just a patch version bump for the frontend.
- [x] Bump the SDK version due to the changes for how docker image versions are pulled


alnoki (changes requested):
- [ ] Bump binary versions in SDK tests
- [ ] Bump SDK version as applicable

alnoki (approved):

PR #602, opened 2025-02-13
feat(ECO-2833): Price display updates

xbtmatt:
# Description

- [x] Display price feed in banner as 0% instead of +0% in green when the delta is exactly 0
- [x] De-couple the delta calculation logic from type conversion
- [x] Add an easier to read `ColoredPriceDisplay` re-usable component to visually mute the non-significant figures in a decimalized-price
~- [x] Use 9 sig figs instead of 8~ Scrapped this because it might confuse users
- [x] Re-use as much logic in `FormattedNumber` as possible

# Testing

### Before

image

### After

image


CRBl69 (changes requested):
Mostly looks good besides the small comment I left :+1:.

Also, it seems there are two changes in this PR:

- Minor price feed styling fixes.
- Color changes for the trade history table.

It is even more confusing as the title of the PR is "price feed component updates" and the screenshots show the trade history table. I'll let this one slide, as this PR might have been made before discussing the separation of PRs, but things like this (mismatch between title and description/screenshots) should be an obvious red flag to separate something into two PRs in the future.

CRBl69 (approved):

xbtmatt (commented):
Things changed since approval:

1. Extracted the Intl formatting logic from `FormattedNumber` to re-use in the colored price display component. Moved it to `format-number-string.ts`
2. Added specific brightness values attached to the color value for `PriceColors`
3. Renamed from `NominalPriceDisplay` to `ColoredPriceDisplay`

PR #605, opened 2025-02-14
test(ECO-2835): Test leaderboard exit status

CRBl69:
# Description

This PR adds E2E tests to test `exited` and `last_exit_0` fields for the arena leaderboard history.

### DIFF
--- a/src/typescript/sdk/tests/e2e/arena/general/arena.test.ts
+++ b/src/typescript/sdk/tests/e2e/arena/general/arena.test.ts
@@ -393,14 +393,14 @@ describe("ensures leaderboard history is working", () => {
 
     expect(user1LeaderboardData.exited).toEqual(true);
     expect(user1LeaderboardData.lastExit0).toEqual(true);
-    expect(Number(leaderboard![0].profits) / Number(leaderboard![0].losses)).toBeGreaterThan(1);
+    expect(Number(user1LeaderboardData.profits) / Number(user1LeaderboardData.losses)).toBeGreaterThan(1);
 
     expect(user2LeaderboardData.exited).toEqual(true);
     expect(user2LeaderboardData.lastExit0).toEqual(false);
 
     expect(user3LeaderboardData.exited).toEqual(false);
     expect(user3LeaderboardData.lastExit0).toBeNull();
-    expect(Number(leaderboard![2].profits) / Number(leaderboard![2].losses)).toBeLessThan(1);
+    expect(Number(user3LeaderboardData.profits) / Number(user3LeaderboardData.losses)).toBeLessThan(1);
   }, 15000);
 
   it("verifies the data during a melee with no activity", async () => {
--- a/src/typescript/sdk/tests/e2e/arena/general/arena.test.ts
+++ b/src/typescript/sdk/tests/e2e/arena/general/arena.test.ts
@@ -496,4 +496,88 @@ describe("ensures leaderboard history is working", () => {
     expect(leaderboard3.exited).toEqual(false);
     expect(leaderboard3.lastExit0).toBeNull();
   }, 15000);
+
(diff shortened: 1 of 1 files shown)


xbtmatt (commented):
Yo, thought this was ready so I accidentally rebased it for you to try to merge it.

Only things I changed were using the helper function (`enterHelper`) and changing `let` to `const`; should be in the commit.

It's still failing and there's a console.log in there, so I imagine you were still in the process of working on it, I'll let you continue on it

xbtmatt (changes requested):
I didn't realize that each test in the leaderboard `describe` block is using the same accounts. This makes it unclear whether the tests would still succeed if run in isolation, since they may be sharing state across tests. They look independent because `account1` `account2` and `account3` are redeclared several times, but they're all using the same accounts (`420`, `421`, `422`), where they have existing leaderboard history after the first test.

Since each test waits for the previous one to complete, the further down you go, the more the state is "contaminated" at the start of the test. Each test should create and use its own fresh accounts with no prior history in the database. This ensures there's no shared state or implicit assumptions, making them fully independent, focused on a single behavior, and able to run in isolation

Relying on previous state from another test substantially increases the risk of getting a false positive and makes failures difficult to debug and understand

When it comes to some of our other tests, it isn't as important to ensure completely isolated state between tests, because we're mostly just testing the behavior of serializing, deserializing and inserting state we get from on-chain.

The leaderboard, however, has its own state that we are creating from our own application/business logic, so it's really important to completely isolate each test, since there's nothing on-chain that's *requiring* us to adhere to certain logic. The tests here should essentially act as sanity checks on the expected final state for our business logic, not just that we can properly serialize/deserialize/fetch data from the database

For example, imagine the first test in the `describe` block adds a user to the leaderboard, which masks a flaw in the database insertion logic in some edge case that a test 5 tests down is supposed to be testing. It would then report a false positive and hide the actual issue

It’s possible these tests aren’t actually relying on each other that much, but we need to be sure for them to be meaningful tests. The safest approach is to refactor them to be completely independent so we can eliminate any unintended dependencies. It will add quite a bit of boilerplate and look like you're repeating yourself over and over, but it will make the tests much more thorough, resilient, and meaningful. I can also help you rewrite some of them!

src/typescript/sdk/tests/e2e/arena/general/arena.test.ts
```diff
@@ -496,4 +496,65 @@ describe("ensures leaderboard history is working", () => {
...
+      .select("*")
+      .eq("melee_id", melee.view.meleeID.toString())
+      .then((r) => r.data)
+      .then((r) => (r === null ? null : r.map(toArenaLeaderboardHistoryModel)));
+
+    expect(leaderboard).not.toBeNull();
+    expect(leaderboard).toHaveLength(2);
+
+    let leaderboard1 = leaderboard!.find((l) => l.user.startsWith("0x420"))!;
+    let leaderboard2 = leaderboard!.find((l) => l.user.startsWith("0x421"))!;
```
xbtmatt, line 525: `leaderboard1` and `leaderboard2` makes it sound like two different leaderboards, not two rows in the leaderboard for different users

Let's rename them to make it clear:

```typescript
    let user1LeaderboardData = leaderboard!.find((l) => l.user.startsWith("0x420"))!;
    let user2LeaderboardData = leaderboard!.find((l) => l.user.startsWith("0x421"))!;
```

(and everywhere else they're used)

xbtmatt (commented):
Aside from the accounts and minor other stuff, this is a really great way to write a test. 😄 

It very explicitly tests a single behavior, describes what it's testing in the test name, and is generally really easy to follow and understand

Once #651 is fixed, just change the `leaderboardX` variable names and add the not `undefined` checks and this is good to go

xbtmatt (approved):

PR #606, opened 2025-02-14
feat(ECO-2036): Add "My emojicoins" page

0xForkh:
# Description
Adding the my emojicoins / portfolio page.
Can be accessed:
- By clicking on sender address in trade history table
- By clicking on wallet address in the nav bar to access own portfolio.

# Note about the pagination
I ended up commenting out the pagination and adding scroll. I think it looks much better, and since we would be doing the pagination ont the client side, I don't see any benefit of having pagination.

I think in terms of UX, the scroll is much better. 

If there is potential performance concerns in the case of a wallet with a lot of coins (1000+), I can implement a form of lazy loading where the rows are only added in the DOM when the scroll reaches their position.

# Todo
- ~~Still need to check why some markets are not returned by the indexer. Seems to be happening for the same coins.~~
- ~~Still need to filter out potential fake tokens using the same name.~~
- ~~Replace logic to fetch specific markets with #623  once it's merged to main~~


src/typescript/frontend/src/lib/aptos-indexer/fungible-assets.ts
```diff
@@ -0,0 +1,115 @@
...
+export async function fetchFungibleAssetsBalance({
+  ownerAddress,
+  assetType,
+  offset = 0,
+  limit = 100,
+}: FetchFungibleAssetsParams): Promise<FetchEmojicoinBalancesResponse> {
+  const conditions = ['{ metadata: {token_standard: {_eq: "v1"}} }', '{ amount: {_gt: "0"} }'];
+
+  if (assetType) conditions.push("{ asset_type: {_eq: $assetType} }");
+  else conditions.push('{ asset_type: {_ilike: "%coin_factory::Emojicoin%"} }');
```
xbtmatt, line 40:
```suggestion
  else conditions.push('{ asset_type: {_ilike: "%::coin_factory::Emojicoin"} }');
```

Are you fetching both the regular `Emojicoin` *and* `EmojicoinLP` coins here? If not, you can make this more efficient and accurate by removing the end `%`, since the coin type should end there

I believe you can also make it slightly more specific with the initial `::`, too

src/typescript/frontend/src/lib/aptos-indexer/fungible-assets.ts
```diff
@@ -0,0 +1,115 @@
...
+    },
+  });
+}
+
+export async function fetchAllFungibleAssetsBalance({
+  ownerAddress,
+  assetType,
+  // Fetch at most 1000 by default, which corresponds to 10 requests.
+  max = 1000,
+}: { max?: number } & Omit<FetchFungibleAssetsParams, "offset" | "limit">): Promise<
```
xbtmatt, line 81: Is this specifically for fetching a single owner's comprehensive FA balances? If so I think you can make the `ownerAddress` param required here just to be more explicit. I believe it's optional from the `FetchFungibleAssetParams` type

src/typescript/frontend/src/app/global.css
```diff
@@ -5,6 +5,7 @@
 @tailwind utilities;
 
 :root {
+  color: white;
```
xbtmatt, line 8: I think this messed with some of the modal stuff that doesn't use a black background by default (see the connect wallet modal)

I don't mind changing the default to white since I frequently find myself having to add `text-white`, but just make sure you comb through everything to make sure nothing is accidentally suddenly white

src/typescript/frontend/src/app/wallet/[address]/page.tsx
```diff
@@ -0,0 +1,111 @@
...
+  // Querying with the emoji directly causes issues for composite emojis such as 🇺🇸 which is a combination of 🇺 and 🇸.
+  const marketAddresses = ownedTokens.flatMap((coin) =>
+    getMarketAddress([...coin.metadata.symbol] as SymbolEmoji[]).toString()
+  );
+
+  // Split addresses into chunks to avoid URL length limits.
+  const chunkSize = 50;
+  const chunks = Array.from({ length: Math.ceil(marketAddresses.length / chunkSize) }, (_, i) =>
+    marketAddresses.slice(i * chunkSize, (i + 1) * chunkSize)
+  );
```
Comment (xbtmatt) [inline src/typescript/frontend/src/app/wallet/[address]/page.tsx:51]: coincidentally, I also added a `chunk` function in #623, in case you want to use that here: https://github.com/econia-labs/emojicoin-dot-fun/blob/d0b331276ef3b5f935cde2991d1614fad15515aa/src/typescript/sdk/src/utils/misc.ts#L251

src/typescript/frontend/src/lib/aptos-indexer/fungible-assets.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  // Filter out non-emojicoin tokens.
+  return allTokens.filter((token) => {
+    // Return early if symbol is not existing emoji.
+    if (!(token.metadata.symbol in SYMBOL_EMOJIS)) return;
+    const address = getEmojicoinMarketAddressAndTypeTags({
+      symbolBytes: encodeEmojis([...token.metadata.symbol]),
+    });
+    return address.emojicoin.toString() === token.asset_type;
```
xbtmatt, line 113:
```suggestion
    return address.emojicoin.toString() === parseTypeTag(token.asset_type).toString();
```

I think you need to standardize the `token.asset_type` here (unless I missed it elsewhere above)

It's basically non-deterministic whether or not you're getting a standardized (no leading zeros) address back from the graphQL endpoint because it's difficult to know whether which "address-like" values are normalized by the TS SDK.

Basically, assume it's *not* standardized, otherwise you'll get a false negative here for things like `0x000000abc...::coin_factory::Emojicoin`

src/typescript/frontend/src/lib/aptos-indexer/fungible-assets.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  // If specific assetType provided, no need to filter out.
+  if (assetType) {
+    return allTokens;
+  }
+
+  // Filter out non-emojicoin tokens.
+  return allTokens.filter((token) => {
+    // Return early if symbol is not existing emoji.
+    if (!(token.metadata.symbol in SYMBOL_EMOJIS)) return;
```
xbtmatt, line 109: So unfortunately the logic for this is a bit more convoluted- the `symbol` here the graphQL response is referring to is the full coin symbol, whereas we refer to an individual symbol emoji as a `symbolEmoji`, and a full symbol is just referred to as a `symbol` (hopefully...it should be consistent)

So for example, a coin symbol on-chain could be:

```typescript
const validSymbol = "⚽✨⚾";
// This will fail, because it's multiple emojis concatenated:
validSymbol in SYMBOL_EMOJIS
// or, equivalently, the map lookup will also fail:
SYMBOL_EMOJI_DATA.hasEmoji(validSymbol)

// To test each emoji you would have to:
const emojis = getSymbolEmojisInString(validSymbol);
// and you'd then know:
expect(emojis).toEqual(["⚽", "✨", "⚾"])
```

The other problem is that `getSymbolEmojisInString(...)` automatically filters out non-symbol emojis (it's just a regex on emojis, filtered by valid symbol emojis), so "asdf⚽" would return "⚽" and thus erreonously succeed in a comparison

The most robust equality/validity check will need to store the initial length of the `token.metadata.symbol`,  call `getSymbolEmojisInString(...)` and join those emojis, then compare the length of the initial length to the length of the joined emojis string

xbtmatt (commented):
This is great, man! Very snappy and I like the way you implemented the faded sort buttons, I wasn't sure how to do that myself with the leaderboard.

Small ux/ui stuff:
1. The layout shifts slightly on hover for each row because there's no border and then suddenly a 1px border, I think you can solve it by making the initial border 1px but transparent, then changing the color on hover
2. (I ran into this same issue) the header row loses its top and bottom border when using `position: sticky`. The only way I was able to fix this was with `before` and `after` pseudo-elements: see [here](https://github.com/econia-labs/emojicoin-dot-fun/blob/997b3289674d961075dec78ff438c3cc93729276/src/typescript/frontend/src/components/pages/stats/HeaderRow.tsx#L27)
3. Add a little more space between the default grayed out up & down indicator and the header text to match the gap between the indicator and the header text when it's active
4. You don't need to truncate the ANS name when it's at the top- I think this is an assumption we made when making the ANS display name component initially- you can refactor it to not truncate on certain conditions? For an example, check this address on mainnet: 0x561178cc72daec681de563a7fb34c61a45ba32058dce54fa8e8b88945430f367
5. You might need to increase the minimum size of the `Connect Wallet` dropdown, because the scramble effect causes a resize due to the characters shifting the layout. Since `My emojicoins` is longer than anything we previously had in there, the scramble now causes a layout shift, so just add a bit of width to it to make sure the scramble doesn't ever increase the width of the dropdown


Overall looks great!! Excited to get it in 😄

src/typescript/frontend/src/lib/aptos-indexer/fungible-assets.ts
xbtmatt: Oooh gotcha, no that's fine, just making sure I understand!

xbtmatt (approved):
Really clean, man, looks great! Excited to get this in

PR #615, opened 2025-02-15
chore(ECO-2847): Remove disabling arena indexing override

alnoki:
# Description

Remove the override to disable arena indexing, thus allowing a cold upgrade with fully-indexed arena data


alnoki:
Closed due to incorrect branch name

PR #620, opened 2025-02-16
[ECO-2636] Create table of coin holders

0xForkh:
First implementation of coin holders table

The base of this branch is #606 , since there is a lot of code in common.
I will rebase this branch on main after #606 is merged.

# Todo
- ~~Improve layout and position on the page~~
- ~~Work on mobile layout~~
- ~~Sticky header~~
- ~~Visual improvements~~


xbtmatt (commented):
Clicking around this and testing it a bit (it's awesome btw)

1. It would be really cool if you could resolve ANS names to their addresses in the URL. It doesn't need to be *super* robust, but a simple query param of `<name>.apt` could resolve the ANS name to their address before fetching their balances. Basically anything that corresponds to `<name>.apt` should be tried as an ANS name
2. Could be totally wrong on this- but might want to double check the balance as % of supply- it looks like it might be off by a factor of 100? Not sure though, it could be accurate

xbtmatt (approved):

PR #637, opened 2025-02-19
feat(ECO-2895): Refresh app data when melee changes [3/4]

xbtmatt:
# Description

Trigger a page refresh to refresh the server component data when a melee changes.

Ideally, the server components work seamlessly and update in real time without a page flash. (No visual refresh on the entire page, only on the individual server components)

So far, this works on the *second* time, but not on the first refresh. I'm not sure if this is an issue with local development or an issue with server components in general. It remains to be seen, but for now this is a simple and effective solution to getting the user to see the latest melee data. 

# Testing

Using a lot in local development- can trigger a new melee and see it update live in front of you, even if you aren't the one that triggered the new melee.

# Checklist


CRBl69:
Actually, I kind of see the problem.

In `arena/page.tsx` (server), we pass `market0` and `market1` to `ArenaClient.tsx` (client). Without refreshing, the market data would be out of date. I don't see a clean solution around this, so I think the way you did it is actually the cleanest. Would be nice to figure out the bug you're describing in the future.

CRBl69 (approved):

PR #638, opened 2025-02-20
fix(ECO-2900) Fill table with empty rows when there is not enough items

0xForkh:
This also includes the change to the header title for "amount" into the emojicoin.

image


xbtmatt (commented):
Awesome, thanks for doing this 😀 

If you could somehow remove the extra borders on non-placeholder rows that'd be awesome, but if not we can merge this as is and I can fix it later

You can see the difference along the edge here:

image


Also if you could add the box shadow on hover that'd be great:

image

image

But I know it had a weird initial implementation so I can also add that back too if you can't get around to it by tomorrow

Actually after posting that I now realize I think you're using a slightly lighter gray here- I believe it should be the color `SENDER` is

xbtmatt (approved):

PR #640, opened 2025-02-20
feat(ECO-2876): Add arena candlesticks to SDK

CRBl69:
# Description

This PR updates the SDK with the new candlestick types.

It adds types for the `arena_candlestick` table.


xbtmatt (commented):
I would be extra safe with this and not include it in any of the existing Period stuff, neither types nor the enum. In it should be typed differently than the other periods, because it's not a contract type. It will make it fairly annoying to work with, but if it's included in the existing types, TypeScript won't warn us about some functions breaking (that should break)

They're just used in so many different places and there are assumptions in a few places on the number of periods. (check `calculatePeriodBoundariesCrossed`, and in general a few of the utility/conversion functions you missed below)

Honestly, it might be a little difficult to properly add it in, so I'm okay if you remove them entirely. Although they would be really nice to have, especially for testing, I think they might be a little too difficult to add properly.

I apologize for suggesting we use them without fully thinking it through, there's just so many places you'd need to update it to get it working. It won't fail until you run the frontend, either...

We can discuss on call tomorrow, too

xbtmatt (commented):
I think this needs to be updated, right? The candlesticks aren't nullable anymore

src/typescript/sdk/src/indexer-v2/json-bigint.ts
```diff
@@ -25,7 +25,7 @@ const tryWithFallbackParse = (parser: (v: any) => any) => (v: any) => {
     return v;
   }
 };
-const parseFloat = (v: any) => Big(v).toString();
+const parseFloat = (v: any) => (v === null ? null : Big(v).toString());
```
xbtmatt, line 28:
```suggestion
const parseFloat = (v: any) => Big(v).toString();
```

Since the candlestick prices are no longer `Nullable`.

xbtmatt (approved):

alnoki (approved):

xbtmatt (approved):

PR #643, opened 2025-02-21
fix(ECO-2907): Non-visual my emojicoin and top holders fixes

xbtmatt:
# Description

- [x] Fix critical error by converting `AccountAddress` to string before sending to graphql
- [x] Update function names, stylistic changes, and move functions to the SDK to properly expose them for future use
- [x] Reduce `revalidate` time for top holders from `60` to `10`
- [x] Update input types to indicate the type is a struct tag string, not just any string
- [x] Add documentation
- [x] Make function names more explicit, move to SDK

# Testing

- [x] My Emojicoin page should no longer be broken
- [x] Revalidate time in local cache should be 60, not 10
- [x] Added tests for parsing functions

# Checklist


src/typescript/frontend/src/components/pages/emojicoin/components/trade-history/trade-history.tsx
```diff
@@ -138,7 +140,7 @@ export const TradeHistory = (props: TradeHistoryProps) => {
       }
       textFormat="body-sm"
       columns={columns}
-      getKey={(item) => item.version.toString()}
+      getKey={(item) => item.guid}
```
xbtmatt, line 143: Otherwise results in a duplicate key error in React

PR #645, opened 2025-02-22
fix(ECO-2910): Export the broker client and indexer types properly to the SDK

xbtmatt:
# Description

- [x] Export the broker client and indexer types properly to the SDK
- [x] Publish to npm and test to see it works as expected
- [x] Change version to `0.5.2` and publish once this PR is merged to `main`
- [x] Fix any residual type errors
- [x] Fix the `EmojicoinClient` class bound function types not being exposed properly by letting TypeScript infer them

# Testing

Tested it while writing the README.md for the SDK in #645

# Checklist

### DIFF
--- a/src/typescript/sdk/package.json
+++ b/src/typescript/sdk/package.json
@@ -117,5 +117,5 @@
     "verify-processor-data": "pnpm dotenv -e ../.env -- pnpm tsx --conditions=react-server src/scripts/verify-processor-data.ts"
   },
   "typings": "dist/common/index.d.ts",
-  "version": "0.5.1"
+  "version": "0.5.2"
 }
--- a/src/typescript/sdk/package.json
+++ b/src/typescript/sdk/package.json
@@ -117,5 +117,5 @@
     "verify-processor-data": "pnpm dotenv -e ../.env -- pnpm tsx --conditions=react-server src/scripts/verify-processor-data.ts"
   },
   "typings": "dist/common/index.d.ts",
-  "version": "0.5.0"
+  "version": "0.5.1"
 }
--- a/src/typescript/sdk/package.json
+++ b/src/typescript/sdk/package.json
@@ -117,5 +117,5 @@
     "verify-processor-data": "pnpm dotenv -e ../.env -- pnpm tsx --conditions=react-server src/scripts/verify-processor-data.ts"
   },
   "typings": "dist/common/index.d.ts",
-  "version": "0.5.2"
+  "version": "0.5.2-rc.1"
 }
--- a/src/typescript/sdk/package.json
+++ b/src/typescript/sdk/package.json
@@ -117,5 +117,5 @@
     "verify-processor-data": "pnpm dotenv -e ../.env -- pnpm tsx --conditions=react-server src/scripts/verify-processor-data.ts"
   },
   "typings": "dist/common/index.d.ts",
-  "version": "0.5.2-rc.1"
+  "version": "0.5.2"
 }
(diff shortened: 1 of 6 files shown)


src/typescript/sdk/src/client/emojicoin-client.ts
```diff
@@ -111,12 +111,7 @@ export class EmojicoinClient {
     remove: this.removeLiquidity.bind(this),
   };
 
-  public utils: {
-    emojisToHexStrings: typeof EmojicoinClient.prototype.emojisToHexStrings;
-    emojisToHexSymbol: typeof EmojicoinClient.prototype.emojisToHexSymbol;
-    getEmojicoinInfo: typeof EmojicoinClient.prototype.getEmojicoinInfo;
-    getTransactionEventData: typeof EmojicoinClient.prototype.getTransactionEventData;
-  } = {
```
xbtmatt, line 119: Without this, the types are opaque to typescript, and it doesn't properly infer, so `client.emojisToHexStrings(...)` has an input arg and return type of `any`. Removing it lets TypeScript properly infer the types and works as expected

src/typescript/sdk/src/indexer-v2/types/index.ts
```diff
@@ -508,7 +509,7 @@ export const withArenaVaultBalanceUpdateData = curryToNamedType(
   "arenaVaultBalanceUpdate"
 );
 
-const EVENT_NAMES = {
```
xbtmatt, line 511: Removing duplicate `EventName` type by importing the previously existing one. They were otherwise exactly the same

src/typescript/sdk/src/indexer-v2/types/index.ts
```diff
@@ -542,7 +541,7 @@ export const GuidGetters = {
     const registryNonce = "registry_nonce" in data ? data.registry_nonce : data.registryNonce;
     return {
       eventName,
-      guid: `${eventName}::${registryNonce}::` as const,
```
xbtmatt, line 545: Removing accidental trailing semicolons. This is used on a per-application client-side basis and is only really for deduplicating events, so it doesn't make any substantial difference. Even if it was already stored  in local storage, we'd just get duplicate global state events, but we currently don't even use them anywhere.

src/typescript/sdk/src/index.ts
```diff
@@ -1,6 +1,8 @@
 export * from "./const";
+export * from "./broker-v2";
 export * from "./emoji_data";
 export * from "./emojicoin_dot_fun";
+export * from "./indexer-v2/types";
```
xbtmatt, line 5: Add the broker and indexer-v2 types to the root export for the SDK.

This lets developers import indexer types without having to import `@econia-labs/emojicoin-sdk/indexer-v2`, making it much less likely for them to run into a `Error: A client component imported the "server-only" module` conflict

src/typescript/sdk/package.json
```diff
@@ -117,5 +117,5 @@
     "verify-processor-data": "pnpm dotenv -e ../.env -- pnpm tsx --conditions=react-server src/scripts/verify-processor-data.ts"
   },
   "typings": "dist/common/index.d.ts",
-  "version": "0.5.0"
+  "version": "0.5.2"
```
CRBl69, line 120: Is there a reason we skip a version here ?

CRBl69 (approved):
Thanks !

Mostly good so I'll approve.

Only question I have is why we skip `0.5.2`.

src/typescript/sdk/package.json
xbtmatt: Yeah my bad, meant to explain this: I accidentally tagged things incorrectly while testing it, so 0.5.1 doesn't properly export everything. Probably need a github action to publish to npm for us

PR #650, opened 2025-02-23
feat(ECO-2913): Add a `scripts` package to the turbo monorepo; add arena scripts

xbtmatt:
# Description

Add a `scripts` package to the TypeScript monorepo! This allows us to test locally and even on testnet.

Added READMEs and explanations for things. This code in general will not be held to the same standard as `src/sdk` and `src/frontend` code.

The intention is simply to have a place to add easily runnable test scripts and consume the `@econia-labs/emojicoin-sdk` package as an external developer.

# Testing

Tested it myself. Check the `README.md` and `scripts/package.json` for simple scripts to run.

# Checklist


src/typescript/scripts/tsconfig.json
```diff
@@ -0,0 +1,50 @@
...
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "outDir": "./dist",
+    "paths": {
+      "@/contract-apis": [
+        "node_modules/@econia-labs/emojicoin-sdk/src/emojicoin_dot_fun/contract-apis"
+      ],
+      "@/contract-apis/*": [
+        "node_modules/@econia-labs/emojicoin-sdk/src/emojicoin_dot_fun/contract-apis/*"
+      ]
```
xbtmatt, line 31: Re-alias paths used in the manually imported test utility functions to point at the dependency in `node_modules`, not to the one in `../sdk/`.

This is slightly hacky, but the alternative is exporting test utility functions from the SDK, which I don't think we want to do.

src/typescript/scripts/src/test-exports/arena-utils.ts
```diff
@@ -0,0 +1,38 @@
+import {
+  registerAndUnlockInitialMarketsForArenaTest,
+  waitUntilCurrentMeleeEnds,
+  setNextMeleeDurationAndEnsureCrank,
+  depositToVault,
+  ONE_MINUTE_MICROSECONDS,
+  ONE_SECOND_MICROSECONDS,
+} from "../../../sdk/tests/e2e/arena/utils";
+import { getPublisher } from "../../../sdk/tests/utils";
```
CRBl69, line 9: Could you export these (in a dedicated PR) ? I can see someone using `waitUntilCurrentMeleeEnds` for example, or `depositToVault`.

src/typescript/scripts/tsconfig.json
CRBl69: I just saw this comment after making the previous one. Is there a reason we don't want to export them. If we need them, it is highly probable that other people will too.

src/typescript/scripts/src/test-exports/arena-utils.ts
alnoki: @CRBl69 I suggest you wrap this into ECO-2915

src/typescript/scripts/tsconfig.json
xbtmatt: I guess I was taking a conservative approach with exporting things from the SDK because it can get confusing when some functions just assume you're running locally. 

For example, if you see a function exported like `setNextMeleeDurationAndEnsureCrank` you might try to call that on `testnet` or `mainnet` but it has the localnet publisher private key hard-coded in them

src/typescript/scripts/src/test-exports/arena-utils.ts
xbtmatt: Yes, as long as we just remove the assumptions about being on a local network specifically it'd help to have these exported

In general it'd be great to clean up the test utility functions and export them

src/typescript/scripts/src/arena/local/bots.ts
```diff
@@ -0,0 +1,49 @@
...
+  const melee = await fetchAllCurrentMeleeData();
+
+  const results = await Promise.allSettled(
+    accounts.map((account) =>
+      makeRandomTrades({
+        melee,
+        account,
+        numTrades: NUM_TRADES,
+      }).catch(async (e) => {
+        // Most likely, a new melee has started. Refetch and reset it.
```
xbtmatt, line 29:
```suggestion
        // Most likely, this error message means a new melee has started.
```

src/typescript/scripts/src/arena/local/bots.ts
```diff
@@ -0,0 +1,49 @@
...
+  const melee = await fetchAllCurrentMeleeData();
+
+  const results = await Promise.allSettled(
+    accounts.map((account) =>
+      makeRandomTrades({
+        melee,
+        account,
+        numTrades: NUM_TRADES,
+      }).catch(async (e) => {
+        // Most likely, this error message means a new melee has started.
```
xbtmatt, line 29:
```suggestion
        // Unless the initial types are somehow incorrect (unlikely), this error
        // message means a new melee has started.
```

src/typescript/scripts/src/arena/local/bots.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const results = await Promise.allSettled(
+    accounts.map((account) =>
+      makeRandomTrades({
+        melee,
+        account,
+        numTrades: NUM_TRADES,
+      }).catch(async (e) => {
+        // Unless the initial types are somehow incorrect (unlikely), this error
+        // message means a new melee has started.
```
xbtmatt, line 30:
```suggestion
        // message means a new melee has started because the types in `melee`
        // are now stale.
```

PR #661, opened 2025-02-28
feat(ECO-2921): Wallet trade history

0xForkh:
## Done
- Add API route to fetch transactions by sender and marketId
- Add infinite scroll and other minor improvements to EcTable
- Add user trade history on portfolio page
- Add shadcn Tabs component (customized with EC style)
- Add my trades history on market page
- Add horizontal scroll for tabs on mobile when overflowing (market page)
- Test and Improve layout on mobile
- Change swapper to sender for market trade-history
- Add filter by marketId/emojicoin for the trade history on portfolio page
- Add sorting for the trade history (recent trades first, or oldest trades first)

### DIFF
--- a/src/typescript/frontend/src/app/wallet/[address]/page.tsx
+++ b/src/typescript/frontend/src/app/wallet/[address]/page.tsx
@@ -4,8 +4,7 @@ import { getAptPrice } from "lib/queries/get-apt-price";
 import { type Metadata } from "next";
 
 export const metadata: Metadata = {
-  title: "Explore the cult",
-  description: `Explore the emojicoin cult`,
+  title: "Wallet Portfolio",
 };
 
 export default async function WalletPage({ params }: { params: { address: string } }) {


src/typescript/frontend/src/components/ui/table/ecTableBody.tsx
```diff
@@ -28,6 +32,19 @@ export const EcTableBody = <T,>({
...
   }, [rowHeight, containerHeight]);
 
+  if (items.length === 0)
+    return (
+      <TableBody className={className}>
+        <tr style={{ height: containerHeight - rowHeight - 1.5 + "px" }}>
+          <TableCell colSpan={columns.length}>
+            <div className="flex justify-center items-center">
+              {isLoading ? <AnimatedLoadingBoxes numSquares={11} /> : "Empty"}
+            </div>
```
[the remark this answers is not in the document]
xbtmatt: Oooh I see what you're saying. Hmm good point. Would it be difficult to fade it (the empty rows) and overlay a "No trade history" where empty is at? Just so it's not as visually jarring to go from one table to the next

src/typescript/frontend/src/components/pages/wallet/useSwapEventsQuery.ts
```diff
@@ -0,0 +1,25 @@
...
+import { type z } from "zod";
+
+export type SwapEvent = Awaited<ReturnType<typeof fetchSwapEvents>>[number];
+
+export const useSwapEventsQuery = (args: z.input<typeof GetTradesSchema>, disabled?: boolean) => {
+  const query = useInfiniteQuery({
+    queryKey: ["fetchSwapEvents", args],
+    queryFn: ({ pageParam }) =>
+      axiosInstance
+        .get<SwapEvent[]>("/api/trades", { params: { ...args, page: pageParam } })
```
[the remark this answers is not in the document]
xbtmatt: also after looking into it it looks like `axios` can interfere with nextjs' implementation of `fetch`, specifically with middleware and cache revalidation. I don't think it matters here since you're diong client side react query stuff, but let's keep it simple and continue with `fetch` to avoid using multiple implementations and getting unexpected behavior in some places

I think the better solution is probably just extending `fetch` to call `parseJSON` and returning the generic passed in

xbtmatt (commented):
Nice, man looks great. Thanks for the small fixes and whatnot. Mainly before getting it in we just need to:

- Rename `marketId` => `marketID` everywhere
- Keep using native (or more accurately, nextjs' extended) `fetch`
- Add new route to `ROUTES` in `routes.ts`

I tagged them all in their corresponding places as well

xbtmatt (commented):
Did you mean to bump the trading view chart library? I think they updated and you pulled in the latest changes.

Not sure what they are so don't want anything to break if it's not backwards compatible

src/typescript/frontend/src/app/api/trades/schema.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const GetTradesSchema = z.object({
+  sender: z
+    .string()
+    .refine((arg) => AccountAddress.isValid({ input: arg }))
+    .optional()
+    .transform((val) => (val ? AccountAddress.from(val) : undefined)),
+  page: z.coerce.number().default(1),
+  limit: z.coerce.number().max(100).default(100),
+  marketID: z.string().optional(),
```
xbtmatt, line 14:
```suggestion
  page: Schemas["PositiveInteger"].default(1),
  limit: Schemas["PositiveInteger"].pipe(z.coerce.number().max(100).default(100)),
  marketID: Schemas["PositiveInteger"].optional(),
```

src/typescript/frontend/src/app/api/trades/schema.ts
```diff
@@ -0,0 +1,24 @@
+import { AccountAddress } from "@aptos-labs/ts-sdk";
+import { isValidEmojiHex, symbolBytesToEmojis } from "@sdk/emoji_data";
+import { toOrderBy } from "@sdk/indexer-v2/const";
+import { z } from "zod";
```
xbtmatt, line 4:
```suggestion
import { toOrderBy } from "@sdk/indexer-v2/const";
import { Schemas } from "@sdk/utils/validation";
import { z } from "zod";
```

src/typescript/frontend/src/lib/ecFetch/ecFetch.tsx
```diff
@@ -0,0 +1,154 @@
...
+/**
+ * Enhanced fetch wrapper with built-in JSON handling and search params support
+ * @template T - Response type from ResponseTypeMap
+ * @param {string | URL | Request} input - URL or Request object
+ * @param {EcFetchInit} [init] - Request configuration
+ * @returns {Promise<T>} Parsed response data
+ * @throws {EcFetchError} If the request fails or response cannot be parsed
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const ecFetch = async <T = any,>(
```
xbtmatt, line 123:
```suggestion
export const ecFetch = async <T = unknown,>(
```

xbtmatt (approved):

src/typescript/frontend/src/lib/ecFetch/ecFetch.tsx
CRBl69, line 1: This should be a `.ts` file

CRBl69, line 1: You need to make sure this works with different revalidates.

Basically, we had problems where you had a `export const getData() = () => fetch(..., {revalidate: 10})` in `lib.ts`.

Then, we had

```tsx
// a/page.tsx

export const revalidate = 1;

export default async function () {
    const data = await getData();
    return <div>{JSON.stringify(data)}</div>;
}
```

and

```tsx
// b/page.tsx

export const revalidate = 60;

export default async function () {
    const data = await getData();
    return <div>{JSON.stringify(data)}</div>;
}
```

and the revalidate from `{a,b}/page.tsx` would overwrite the revalidate from the fetch call. So the data we wanted cached for exactly 10 seconds would get either cached for 1 or 60 seconds.

We had **A LOT** of headaches resolving this and I'm very hesitant allowing changes to the caching mechanism into the repo.

Could you make sure this is tested ?

Ideally in a E2E test, but if that is too long, at least a well conducted manual test with some screenshots as a reply to this comment would be appreciated.

Thanks.

CRBl69: Also test this on pages where `dynamic = 'force-dynamic'` or `revalidate = 0` is set.

xbtmatt: We did change `experimental.staleTimes.dynamic` from `30` to `0` recently (in `next.config.mjs`, and that was a really big part of the issue. This is the flag they specifically turned off by default in nextjs 15.

So it's possible our understanding of the caching mechanism was actually incorrect and there was no issue aside from the stale times (which is more of a browser/router cache thing).

xbtmatt (changes requested):
I really do appreciate the inspiration and desire to create a more ergonomic `fetch` that handles all of our specific needs, but after re-reviewing this, I think the scope of what this is trying to accomplish is too large for one PR and will likely continue to balloon if the fetch/cache stuff is in here

I think this PR should just focus on completing the wallet trade history and unifying event types to all have the `sender` field. Keep in mind that lots of event types have changed or been added since the PR was opened

---

As for the new `fetch` stuff, I think you should keep it as simple as possible. I'm pretty sure we exclusively make `GET` requests (no `POST`s except for server actions, and we shouldn't really be using those anyway).

We also don't use a `Request` anywhere in the codebase currently, and the construction of search params is a little clunky, but it's not very ahrd to do.

I really do appreciate you trying to incorporate all possible paths,  but we should keep any extensions to fetching mechanisms as simple as possible to avoid introducing cognitive overhead and potential failure paths that may not be exposed until later (since we don't use a lot of them).

Instead of a wrapper for `fetch` I think it would make more sense to just have helper functions dedicated to the specific use cases we need. 

I think initially you wrote this to emulate axios's behavior of throwing errors for non-2xx status codes in a response, which is just 2-3 lines of code. I don't know if any of the other stuff is necessary, and I think converting individual fetch functions to use `revalidate` can be an entirely separate PR dedicated to refactoring *all* native `fetch` calls.

Keep in mind we use 3rd party libraries to fetch data pretty frequently (`postgrest` and `aptos` fullnode/indexer queries), so a lot of the stuff here can't be re-used by those.

I think the highest priorities for improving our `fetch` implementations should just be addressing the pitfalls and issues shared across native fetch and 3rd party fetches (postgrest, aptos, etc), like serializing/deserializing and caching data. This can be done by:
  - Improving the ergonomics of `unstable_cache`, possibly even just by properly testing it and documenting its behavior
  - Unifying parsing/stringification functions everywhere
  - Adding proper error handling for non-2xx responses
  
Currently, there's nothing *broken* about our current implementation, so I don't think it's worth the additional complexity and cognitive overhead of adding a new way of fetching data. What you wrote is a good start, but I think ultimately if we add something like this, it should replace every other instance of `fetch` in the codebase, so there's very little cognitive overhead to understanding what a specific `fetch` is doing under the hood

xbtmatt (approved):
Looks good man, just a few small things!

alnoki (approved):

src/typescript/frontend/src/lib/ecFetch/ecFetch.tsx
xbtmatt: Resolving because the file was removed

xbtmatt: Resolving because `ecFetch` and the cache stuff in general was moved to another PR. Although `get-apt-price` does use the native `fetch` with `revalidate`, it was thoroughly tested and works well.

xbtmatt (approved):

PR #668, opened 2025-03-03
fix(ECO-2944): Standardize account addresses in arena queries; fix import

xbtmatt:
# Description

Very straightforward- just a small fix to standardize addresses and fix an import.


CRBl69 (approved):
LGTM, thanks

PR #675, opened 2025-03-06
feat(ECO-2984): Add my favorites Move package

CRBl69:
# Description

This PR adds a move package to save and get favorite user markets on chain.

# Testing

100% test coverage.

# Checklist


src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,251 @@
...
+    const E_MARKET_NOT_FOUND: u64 = 1;
+
+    struct Nil {}
+    has copy, drop, store;
+
+    struct FavoriteData {
+        market: address,
+        favorites: u128
+    }
+    has copy, drop, store;
```
alnoki, line 21: Why is this here? It isn't used

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,251 @@
...
+
+    public entry fun set_favorite(user: &signer, market: address) acquires Registry {
+        assert!(
+            option::is_some(&emojicoin_dot_fun::market_metadata_by_market_address(market)),
+            E_MARKET_NOT_FOUND
+        );
+        let user_address = signer::address_of(user);
+        let vault_ref_mut = &mut Registry[@favorites];
+        if (vault_ref_mut.users.contains(user_address)) {
+            vault_ref_mut.users.borrow_mut(user_address).upsert(market, Nil {});
```
alnoki, line 44: Why did you use the word vault? Is this copy-pasted?

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,251 @@
...
+    const ACCOUNT_2: address = @0x2222;
+    #[test_only]
+    const MARKET_1: vector<vector<u8>> = vector[x"f09f9088e2808de2ac9b"];
+    #[test_only]
+    const MARKET_2: vector<vector<u8>> = vector[x"f09f96a4"];
+
+    #[test_only]
+    const ESIMPLE_MAP_NOT_FOUND: u64 = 65538;
+    #[test_only]
+    const ESMART_TABLE_NOT_FOUND: u64 = 65537;
```
alnoki, line 84: For consistency use `E_ERROR_CODE...`

Also, what do these error codes mean? Where are they raised?

CRBl69: This are codes from simple map and smart table. They're raised whenever they don't find the requested element.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,251 @@
...
+
+    struct FavoriteData {
+        market: address,
+        favorites: u128
+    }
+    has copy, drop, store;
+
+    struct Registry has key {
+        users: SmartTable<address, SimpleMap<address, Nil>>
+    }
```
alnoki, line 25: Note that the SmartTable can be DoS'ed when it is a global resource, e.g. a malicious actor can upsert arbitrary data and block bucketing

So if it would be a global table, I think it would be better to just use `TableWithLength`, which does not have the same attack vector

But then in that case, each user is going to get allocated their own storage slot, which begs the question of why not just storing each user's data under their account - is it because you want it to be iterable?

I was thinking that it might be easier to just do it as a per-user resource, e.g. `MyFavorites`, that way a view function can just borrow it from the user's address - however then in this case it wouldn't be iterable

If global you could use the new BTreeMap (`BigOrderedTable`), but then I'm still not sure because the value in the BTreeMap (here a simplemap) is of variable size and someone could do grieving and stuff in a huge resource

So my first iteration of redesign goes like this:

`MyFavorites` for each user, which has a smart table of all their favorites. Since an address is a `u256`, you can only store 16 in a storage slot before having to bucket it out without paying a ton of gas (storage slots above 4k get expensive), and thus would have to paginate. So here you'd use `target_bucket_size` in the `new_with_config` api

Then my second iteration of though redesign goes like this:

But actually, I think even better would be looking up by market ID if possible since that is only a `u64` - on the frontend you should already have a map from ID to address, and it's easy enough to verify valid market ID on insert. So here you could theoretically do up to ~64 favorites, but I say we cap it at 32 for overhead buffer and stick with an `OrderedMap<market_id: u64, Nil>`. This way you can get the whole thing out in one view without pagination and ideally won't be too costly

Then, if you want to do a global iterable list of everyone that has favorties, you could stuff a `BigOrderedMap<user: address, n_favorites: u8>` in a registry, which will be iterable and also fixed size so you can use the `big_ordered_map::new` API

Stream of consciousness, I know, but I think it'll be valuable to trace this all out to see how to best last on a design to minimize gas costs and maximimize features. So to summarize:

```rust
/// Global map from user address to number of favorites.
struct Registry has key {
    users: BigOrderedMap<address, u8>
}

/// Set of a user's favorites.
struct Favorites has key {
    /// Maximum of 32 favorites.
    set: OrderedMap<u64, Nil>
}
```

xbtmatt: BTW I'm pretty sure this is a classic top-k problem, where `k` here is just the #  of favorite markets. It should be solvable with a min heap in O(logk) time, which is still feasible even for up to 100,000 markets.

The main tradeoff is that it arbitrarily evicts markets with the same # of favorites, but I don't think that matters because postgres would also do that (non-deterministically removes rows when their sort keys are equal)

alnoki: > It should be solvable with a min heap 

Are you aware of an existing Move structure that supports this? Not sure if bespoke custom is worth it

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: This has been addressed in 1bb1d931. There is no more storage under the package's account.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: This has been addressed in 1bb1d931 by adding `_` between `E` and the rest of the error name.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: The above has been implemented in 1bb1d931.

A `FavoriteData` struct has been created and stored at the user level.

A `Favorite` event is emitted when a user favorites or unfavorites a market.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: This is now used as the struct to store data into a user's account (updated in 1bb1d93).

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,247 @@
...
+    /// Market not found.
+    const E_MARKET_NOT_FOUND: u64 = 1;
+
+    struct Nil {}
+    has copy, drop, store;
+
+    struct FavoriteData {
+        markets: SimpleMap<address, Nil>
+    }
+    has copy, drop, store, key;
```
alnoki, line 26:
```suggestion
    struct FavoriteData has key {
        markets: SimpleMap<address, Nil>
    }
```

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,247 @@
...
+        user: address,
+        market: address,
+        is_favorite: bool
+    }
+
+    /// Market not found.
+    const E_MARKET_NOT_FOUND: u64 = 1;
+
+    struct Nil {}
+    has copy, drop, store;
```
alnoki, line 21:
```suggestion
    struct Nil has copy, drop, store {};
```

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,247 @@
...
+            option::is_some(&emojicoin_dot_fun::market_metadata_by_market_address(market)),
+            E_MARKET_NOT_FOUND
+        );
+        let user_address = signer::address_of(user);
+        event::emit(
+            Favorite { user: user_address, market, is_favorite: true }
+        );
+        if (exists<FavoriteData>(user_address)) {
+            let favorites = borrow_global_mut<FavoriteData>(user_address);
+            favorites.markets.add(market, Nil {});
```
alnoki, line 49: Should this be an upsert?

Or maybe early return if it is already there?

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,247 @@
...
+    public fun favorites(user: address): vector<address> acquires FavoriteData {
+        if (exists<FavoriteData>(user)) {
+            let favorites = borrow_global_mut<FavoriteData>(user);
+            favorites.markets.keys()
+        } else {
+            vector::empty()
+        }
+    }
+
+    public entry fun set_favorite(user: &signer, market: address) acquires FavoriteData {
```
alnoki, line 38:
```suggestion
    /// Set the market as a user's favorite, returning early if already set.
    public entry fun set_favorite(user: &signer, market: address) acquires FavoriteData {
```

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,247 @@
...
+        assert!(
+            option::is_some(&emojicoin_dot_fun::market_metadata_by_market_address(market)),
+            E_MARKET_NOT_FOUND
+        );
+        let user_address = signer::address_of(user);
+        event::emit(
+            Favorite { user: user_address, market, is_favorite: true }
+        );
```
alnoki, line 46: If the user already has the market as a favorite, should it be re-emmitted? I suggest putting this at the end, and early returning if market is already a favorite

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+
+    public entry fun unset_favorite(user: &signer, market: address) acquires FavoriteData {
+        let user_address = signer::address_of(user);
+        event::emit(
+            Favorite { user: user_address, market, is_favorite: false }
+        );
+        let favorites = borrow_global_mut<FavoriteData>(user_address);
+        favorites.markets.remove(&market);
```
alnoki, line 64: How about a silent early return if already is not a favorite?

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,247 @@
...
+            let map = simple_map::new();
+            map.add(market, Nil {});
+            let favorites = FavoriteData { markets: map };
+            move_to<FavoriteData>(user, favorites);
+        }
+    }
+
+    public entry fun unset_favorite(user: &signer, market: address) acquires FavoriteData {
+        let user_address = signer::address_of(user);
+        event::emit(
```
alnoki, line 60: How about emitting this at the end, only after checking that it is already not in the user's favorites?

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,247 @@
...
+            favorites.markets.add(market, Nil {});
+        } else {
+            let map = simple_map::new();
+            map.add(market, Nil {});
+            let favorites = FavoriteData { markets: map };
+            move_to<FavoriteData>(user, favorites);
+        }
+    }
+
+    public entry fun unset_favorite(user: &signer, market: address) acquires FavoriteData {
```
alnoki, line 58:
```suggestion
    /// Remove the market from the user's favorites, returning early if it is not listed.
    public entry fun unset_favorite(user: &signer, market: address) acquires FavoriteData {
```

alnoki (changes requested):

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: I don't want to emit a fav event if present. I also don't want to add a check, in order to reduce the number of instructions used. This will already error out if the element is already present, which will prevent the event being emitted. Also, the caller should never try to favorite an already favorited market, so this seems natural to fail.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: See my previous comment.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: As above, I don't want a useless check. It adds code, and it adds instructions. The `remove` function already performs the check. There is no reason for me to reimplement it.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: See my previous comment.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: Furthermore, I think this lets the user (the frontend dev) know their code does something wrong when trying to fav twice, as opposed to an early return which might let a logic error slip through in the frontend.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: Noting this is copied from `emojicoin_dot_fun_claim_link.move` where it is formatted like this, and running `aptos move fmt` does not change this.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: Addressed in 0488dc9aea5ec350cff047a8fada9038e360dfe9 and 6d3a8a84e5a16e578bcb5f000eed52a076401be7.

Note: I also needed the `drop` ability.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: Added a comment in 0488dc9aea5ec350cff047a8fada9038e360dfe9.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
CRBl69: Added a comment in 0488dc9aea5ec350cff047a8fada9038e360dfe9.

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
alnoki: - [ ] Per discussion add `E_ALREADY_FAVORITE` or similar if already a favorite

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
alnoki: - [ ] Per discussion error with `E_NOT_FAVORITE` if not a favorite

alnoki (changes requested):

alnoki (approved):

src/move/favorites/README.md
```diff
@@ -0,0 +1,69 @@
...
+EMOJICOIN_DOT_FUN=0xbbb...
+PROFILE=my-profile
+```
+
+Publish:
+
+```sh
+NAMED_ADDRESSES=$(
+    printf "%s,%s" \
+        "favorites=$FAVORITES" \
```
xbtmatt, line 22:
```suggestion
        "favorites=$PUBLISHER" \
```

the CLI automatically resolves profiles to their corresponding address when used like this

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,263 @@
...
+    struct Nil {}
+    has copy, drop, store;
+
+    struct FavoriteData {
+        markets: OrderedMap<address, Nil>
+    }
+    has drop, key;
+
+    #[view]
+    public fun favorites(user: address): vector<address> acquires FavoriteData {
```
xbtmatt, line 33:
```suggestion
    public fun view_favorites(user: address): vector<address> acquires FavoriteData {
```

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    /// Set the market as a user's favorite.
+    ///
+    /// An error will be returned if:
+    ///
+    /// - The market does not exist.
+    /// - The market is already a favorite.
+    public entry fun set_favorite(user: &signer, market: address) acquires FavoriteData {
```
xbtmatt, line 48:
```suggestion
    /// Add the market to the user's favorite markets.
    ///
    /// An error will be returned if:
    ///
    /// - The market does not exist.
    /// - The market is already a favorite.
    public entry fun add_favorite(user: &signer, market: address) acquires FavoriteData {
```

Make sure to update in `README.md` too

`set_favorite` is sort of misleading as a name, since it could also mean you're setting a singular favorite to some value or that you're setting the "current" favorite to some value

it's really just adding a favorite to the set of favorited markets

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,263 @@
...
+            move_to<FavoriteData>(user, favorites);
+        }
+    }
+
+    /// Remove the market from the user's favorite.
+    ///
+    /// An error will be returned if:
+    ///
+    /// - The market is not a favorite.
+    public entry fun unset_favorite(user: &signer, market: address) acquires FavoriteData {
```
xbtmatt, line 75:
```suggestion
    public entry fun remove_favorite(user: &signer, market: address) acquires FavoriteData {
```

src/move/favorites/README.md
```diff
@@ -0,0 +1,69 @@
+<!-- cspell:word permissioned -->
+
+# `emojicoin-dot-fun` favorites
+
+This package contains functions to set, get and unset favorite markets per user.
```
xbtmatt, line 5:
```suggestion
This package contains functions to add, remove, and view a user's favorite
emojicoin markets.
```

src/move/favorites/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Get favorites
+
+```sh
+USER_ADDRESS=0xccc...
+FAVORITES=0xaaa...
+PROFILE_NAME=my-profile
+
+aptos move view \
+  --args address:$USER_ADDRESS \
+  --function-id $FAVORITES::emojicoin_dot_fun_favorites::favorites \
```
xbtmatt, line 67:
```suggestion
  --function-id $FAVORITES::emojicoin_dot_fun_favorites::view_favorites \
```

src/move/favorites/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Set favorite
+
+```sh
+MARKET_ADDRESS=0xccc...
+FAVORITES=0xaaa...
+PROFILE_NAME=my-profile
+
+aptos move run \
+  --args address:$MARKET_ADDRESS \
+  --function-id $FAVORITES::emojicoin_dot_fun_favorites::set_favorite \
```
xbtmatt, line 41:
```suggestion
  --function-id $FAVORITES::emojicoin_dot_fun_favorites::add_favorite \
```

src/move/favorites/README.md
```diff
@@ -0,0 +1,69 @@
...
+        "emojicoin_dot_fun=$EMOJICOIN_DOT_FUN"
+)
+aptos move publish \
+    --assume-yes \
+    --move-2 \
+    --named-addresses $NAMED_ADDRESSES \
+    --profile $PROFILE
+```
+
+## Set favorite
```
xbtmatt, line 32:
```suggestion
## Add favorite
```

src/move/favorites/README.md
```diff
@@ -0,0 +1,69 @@
...
+FAVORITES=0xaaa...
+PROFILE_NAME=my-profile
+
+aptos move run \
+  --args address:$MARKET_ADDRESS \
+  --function-id $FAVORITES::emojicoin_dot_fun_favorites::set_favorite \
+  --profile $PROFILE_NAME
+```
+
+## Unset favorite
```
xbtmatt, line 45:
```suggestion
## Remove favorite
```

src/move/favorites/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+```sh
+MARKET_ADDRESS=0xccc...
+FAVORITES=0xaaa...
+PROFILE_NAME=my-profile
+
+aptos move run \
+  --args address:$MARKET_ADDRESS \
+  --function-id $FAVORITES::emojicoin_dot_fun_favorites::unset_favorite \
```
xbtmatt, line 54:
```suggestion
  --function-id $FAVORITES::emojicoin_dot_fun_favorites::remove_favorite \
```

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,263 @@
...
+
+    #[event]
+    struct Favorite has copy, drop, store {
+        user: address,
+        market: address,
+        is_favorite: bool
+    }
+
+    /// Market not found.
+    const E_MARKET_NOT_FOUND: u64 = 1;
```
xbtmatt, line 18:
```suggestion
    /// Market does not exist.
    const E_MARKET_DOES_NOT_EXIST: u64 = 1;
```

xbtmatt (commented):
mostly just stylistic/documentation changes, other than that looks good

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,263 @@
...
+    /// Set the market as a user's favorite.
+    ///
+    /// An error will be returned if:
+    ///
+    /// - The market does not exist.
+    /// - The market is already a favorite.
+    public entry fun set_favorite(user: &signer, market: address) acquires FavoriteData {
+        assert!(
+            option::is_some(&emojicoin_dot_fun::market_metadata_by_market_address(market)),
+            E_MARKET_NOT_FOUND
```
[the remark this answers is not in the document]
CRBl69: I do not see the point of this. It is literally almost the same thing.

src/move/favorites/README.md
```diff
@@ -0,0 +1,69 @@
...
+This package contains functions to set, get and unset favorite markets per user.
+
+## Publish commands
+
+Set variables:
+
+```sh
+FAVORITES=0xaaa...
+EMOJICOIN_DOT_FUN=0xbbb...
+PROFILE=my-profile
```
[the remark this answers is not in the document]
xbtmatt: that's true

xbtmatt (approved):
Updated `favorites` => `view_favorites` since you updated it in the README.md (I took that as you accepting the suggestion but just forgetting to update it in the contract)

Ran tests with it and succeeded, so this lgtm!

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,263 @@
+// cspell:word upsert
+module favorites::emojicoin_dot_fun_favorites {
```
alnoki, line 2:
```suggestion
module favorites::emojicoin_dot_fun_favorites {
```

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,263 @@
...
+            let favorites = borrow_global_mut<FavoriteData>(user);
+            favorites.markets.keys()
+        } else {
+            vector::empty()
+        }
+    }
+
+    /// Add the market to the user's favorite markets.
+    ///
+    /// An error will be returned if:
```
alnoki, line 44:
```suggestion
    /// Aborts if:
```

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -0,0 +1,263 @@
...
+            map.add(market, Nil {});
+            let favorites = FavoriteData { markets: map };
+            move_to<FavoriteData>(user, favorites);
+        };
+        event::emit(Favorite { user: user_address, market, is_favorite: true });
+    }
+
+    /// Remove the market from the user's favorite markets.
+    ///
+    /// An error will be returned if:
```
alnoki, line 72:
```suggestion
    /// Aborts if:
```

src/move/favorites/README.md
```diff
@@ -0,0 +1,70 @@
+<!-- cspell:word permissioned -->
+
+# `emojicoin-dot-fun` favorites
```
alnoki, line 3:
```suggestion
# `emojicoin-dot-fun` favorites
```

alnoki (approved):
Just some minor docs changes now that we've landed on a revision

PR #683, opened 2025-03-07
feat(ECO-2993): Add estimated volume in USD based on current APT price

gregnazario:
# Description

This is just estimated, but it gives users a way to determine what their value trading is with less confusion.

# Testing
Tested locally
Screenshot 2025-03-07 at 6 36 01 PM
Screenshot 2025-03-07 at 6 36 16 PM

# Checklist

### DIFF
--- a/src/typescript/frontend/src/components/pages/emojicoin/components/main-info/MainInfo.tsx
+++ b/src/typescript/frontend/src/components/pages/emojicoin/components/main-info/MainInfo.tsx
@@ -21,7 +21,8 @@ import { useAptos } from "context/wallet-context/AptosContextProvider";
 import type { Colors } from "theme/types";
 import Popup from "components/popup";
 import { DISCORD_METADATA_REQUEST_CHANNEL, LINKS } from "lib/env";
-import { useUsdMarketCap } from "@hooks/use-usd-market-cap";
+import { useUsdMarketCap, useUSDValue } from "@hooks/use-usd-market-cap";
+import { Switcher } from "../../../../switcher";
 
 const statsTextClasses = "uppercase ellipses font-forma text-[24px]";
 
@@ -82,6 +83,7 @@ const MainInfo = ({ data }: MainInfoProps) => {
   const [allTimeVolume, setAllTimeVolume] = useState(
     BigInt(data.state.state.cumulativeStats.quoteVolume)
   );
+  const [showUSD, setShowUSD] = useState<boolean>(false);
 
   useEffect(() => {
     if (stateEvents.length === 0) return;
@@ -96,6 +98,8 @@ const MainInfo = ({ data }: MainInfoProps) => {
   }, [stateEvents]);
 
   const usdMarketCap = useUsdMarketCap(marketCap);
+  const usdDailyVolume = useUSDValue(dailyVolume);
+  const usdAllTimeVolume = useUSDValue(allTimeVolume);
 
   const { isMobile, isTablet } = useMatchBreakpoints();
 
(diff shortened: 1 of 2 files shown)


src/typescript/frontend/src/components/pages/emojicoin/components/main-info/MainInfo.tsx
```diff
@@ -218,6 +220,15 @@ const MainInfo = ({ data }: MainInfoProps) => {
...
                 <AptosIconBlack className="icon-inline mb-[0.3ch]" />
+                {usdDailyVolume !== undefined ? (
+                  <>
+                    &nbsp;
+                    {"(~ "}
+                    <FormattedNumber value={usdDailyVolume} scramble />
+                    &nbsp;
+                    {"$"})
+                  </>
+                ) : null}
```
xbtmatt, line 231: It's a bit crowded visually when the numbers are big like you said: 
image

I think the simplest way to do it here is just render one or the other on an interval of 5 seconds, something like:

```typescript
const NOMINAL_VALUE_INTERVAL = 5000;

// ... in the component
const [showUSD, setShowUSD] = useState(false);
useInterval(() => setShowUSD((v) => !v), NOMINAL_VALUE_INTERVAL);

// in the JSX body
{showUSD ? (
  <>
    <FormattedNumber value={usdDailyVolume} scramble />
    {"$"}
  </>
) : (
  <>
    <FormattedNumber value={dailyVolume} nominalize scramble />
    &nbsp;
    <AptosIconBlack className="icon-inline mb-[0.3ch]" />
  </>
)
```

xbtmatt (commented):
Awesome, thx for this, left a comment below on the USD stuff

Did you mean to include the bridge client page or is that for something separate?

src/typescript/frontend/src/components/pages/emojicoin/components/main-info/MainInfo.tsx
aptosnoob: this is great. i'll suggest make it optional. user should be able to toggle between both options, just like it is on wapal.
Screenshot 2025-03-07 at 21 15 45

gregnazario: Yeah, I'm going to take a look at toggling now.

src/typescript/frontend/src/components/pages/emojicoin/components/main-info/MainInfo.tsx
```diff
@@ -157,6 +161,32 @@ const MainInfo = ({ data }: MainInfoProps) => {
...
+      return (
+        <>
+          <FormattedNumber value={value} nominalize scramble />
+          &nbsp;
+          <AptosIconBlack className={"icon-inline mb-[0.3ch]"} />
+        </>
+      );
+    }
+  }
+
```
xbtmatt, line 189:
```suggestion
  const switcher = <Switcher disabled={valueUsd === undefined} checked={showUSD} onChange={() => setShowUSD((v) => !v)} />;

  function aptOrUsd(value: bigint, valueUsd: number | undefined) {
    if (valueUsd !== undefined && showUSD) {
      return (
        <>
          <FormattedNumber value={valueUsd} scramble />
          {" $"}
        </>
      );
    } else {
      return (
        <>
          <FormattedNumber value={value} nominalize scramble />
          {" "}
          <AptosIconBlack className={"icon-inline mb-[0.3ch]"} />
        </>
      );
    }
  }

```

Make sure to change to `setShowUSD((v) => !v)` bc it avoids issues with closures and stale state. It probably doesn't matter here, but if there are multiple state updates for a single render, the final render might not use the last value unless you do it like this

The other stuff is just to disable the toggle and not worry about displaying it if there's no USD value

src/typescript/frontend/src/components/pages/emojicoin/components/main-info/MainInfo.tsx
```diff
@@ -192,21 +222,19 @@ const MainInfo = ({ data }: MainInfoProps) => {
         </div>
 
         <div className={`flex flex-col justify-between ${borderStyle}`}>
+          <div className="flex justify-between">
+            <div className={statsTextClasses + " text-light-gray"}>
+              APT&nbsp;
+              <AptosIconBlack className={"icon-inline mb-[0.0ch]"} />
+              &nbsp;/&nbsp;USD&nbsp;$:
```
xbtmatt, line 229:
```suggestion
              {"APT "}
              <AptosIconBlack className={"icon-inline mb-[0.0ch]"} />
              {" / USD $:"}
```

Think I got it right, but yea just to make it clearer

xbtmatt (approved):
Awesome, looks great! If you can just make the small changes, this is ready to merge

PR #690, opened 2025-03-10
feat(ECO-2982): Implement site-wide rate limiting

CRBl69:
# Description

This PR adds site-wide rate limitting.

The rate limitting is split up between candlesticks and the rest of the
website, to avoid rate limitting other pages when the user has reached the
candlestick rate limit.

All prefetches have been turned of to avoid running into rate limitting too fast.

Candlesticks fetch has been updated to retry after the rate limitting has ended.

The rate limit is 5 requests per 10 seconds per IP.

# Testing

Spam an endpoint.

# Checklist


src/typescript/frontend/src/middleware.ts
```diff
@@ -21,6 +43,45 @@ export default async function middleware(request: NextRequest) {
...
+
+  const ip = request.ip ?? "127.0.0.1";
+
+  let limit: number;
+  let reset: number;
+  let remaining: number;
+  let domain: string;
+  let success: boolean;
+
+  if (pathname === ROUTES.api.candlesticks || pathname === "/candlesticks") {
```
xbtmatt, line 55: I think this is only the API route now- the `/candlesticks`, is it even possible to query `/candlesticks` still? I would try it and see what the pathname shows up here as, because Said changed a few things with the redirects in next.config.mjs (it's more of a path alias than a 308 redirect now)

xbtmatt (commented):
I think we should try to limit this as much as possible to api routes (and public routes in general) and change the application logic as little as possible.

Things like removing the `prefetch` on all the different markets (and links in general) will get it to work, but that will make navigating the site much slower. nextjs prefetches things prettty intelligently (only does it for things visually in view) so we definitely want to keep it as long as it isn't a huge performance issue. Those fetches are all server side so you should be able to authenticate them somehow or differentiate them between server/client requests

Also the candlesticks retries should just return no rows if rate limited, and then we should make it as lenient as necessary to get candlesticks to fully appear for most markets (even with lots of data)

Also- is `experimental-edge` necessary to run this stuff? Just curious

src/typescript/frontend/src/components/charts/get-bars.ts
```diff
@@ -30,15 +30,38 @@ export const fetchCandlesticksForChart = async ({
...
-    .then((res) => parseJSON<PeriodicStateEventModel[]>(res))
-    .then((res) =>
-      res
-        .sort((a, b) => Number(a.periodicMetadata.startTime - b.periodicMetadata.startTime))
-        .map(toBar)
-        .reduce(curriedBarsReducer(periodParams.to), [])
-    );
+  let data: Bar[] = [];
+  let retriesLeft = 4;
+  while (true) {
```
xbtmatt, line 35: while (retriesLeft-- > 0)

src/typescript/frontend/src/middleware.ts
```diff
@@ -21,6 +34,35 @@ export default async function middleware(request: NextRequest) {
...
+    let limit: number;
+    let reset: number;
+    let remaining: number;
+    let success: boolean;
+
+    const ratelimitRes = await ratelimit.limit(ip);
+    limit = ratelimitRes.limit;
+    reset = ratelimitRes.reset;
+    remaining = ratelimitRes.remaining;
+    success = ratelimitRes.success;
```
xbtmatt, line 50:
```suggestion
    const ratelimitRes = await ratelimit.limit(ip);
    const { limit, reset, remaining, success } = ratelimitRes
```

src/typescript/frontend/src/middleware.ts
```diff
@@ -21,6 +34,35 @@ export default async function middleware(request: NextRequest) {
...
+
+    const headers = {
+      "X-RateLimit-Limit": limit.toString(),
+      "X-RateLimit-Remaining": remaining.toString(),
+      "X-RateLimit-Reset": reset.toString(),
+    };
+
+    if (!success) {
+      return new Response("", {
```
xbtmatt, line 59:
```suggestion
      return new NextResponse("", {
```

just to keep it consistent

xbtmatt (commented):
Nice, added a few suggestions- the main thing is properly implementing the rate limiting for the candlesticks APIs.

I think you should probably just implement a separate rate limiter for the candlesticks APIs (one for both of them), that way we can use a different algorithm.

The token bucket algorithm seems much more suitable, since it's more permissive for flurries of requests, which is what the charting library does that's probably triggering rate limits right now.

### Use two rate limiters

You can separate the two rate limiters with a simple pathname check for `/api/candlesticks` and `/api/arena/candlesticks` vs every other `/api/*` route, and then you just need to add a separate rate limiter for the candlesticks.

Something like this, maybe:

```typescript
const basicRateLimiter = new Ratelimit({
  redis,
  limiter: Ratelimit.slidingWindow(5, "10 s"),
});

const candlestickRateLimiter = new Ratelimit({
  redis,
  limiter: Ratelimit.tokenBucket(10, "30 s", 20),
});

const getRateLimiter = async (pathname: string, ip: string) => {
  const isCandlesticks =
    pathname.startsWith(ROUTES.api.candlesticks) ||
    pathname.startsWith(ROUTES.api.arena.candlesticks);

  const rateLimiter = isCandlesticks
    // `blockUntilReady` lets us move the "try again" logic in the
    // `get-bars.ts` function to middleware, rather than the app.
    ? candlestickRateLimiter.blockUntilReady(ip, 30_000)
    : basicRateLimiter.limit(ip);

  return await rateLimiter.then(({ limit, reset, remaining, success }) => ({
    success,
    headers: {
      "X-RateLimit-Limit": limit.toString(),
      "X-RateLimit-Remaining": remaining.toString(),
      "X-RateLimit-Reset": reset.toString(),
    },
  }));
};
```

### Try the in-memory map, see if it even works

Apparently you can also use an in-memory map? I don't know if this really works though...but it might save us querying the redis instance in case someone is trying to dos the rate limiter itself. Something to consider/try

https://github.com/upstash/ratelimit-js/tree/main/examples/nextjs-middleware

### Disable the rate limiter in non-production environments by default

Don't want to spam the redis instance (or even need the API keys) when run locally.

### Put this all behind a feature flag/env variable

```shell
RATE_LIMITING_ENABLED=1
```

This lets us turn it on and off again really easily without having to go through the whole release cycle. If it's being troublesome or blocking legitimate users, we can disable it while we figure it out.

xbtmatt (approved):

PR #691, opened 2025-03-11
fix(ECO-3018): Catch fetch errors on `[market]/page.tsx`

xbtmatt:
# Description

In general we should be catching errors and providing default values where possible in the higher level server components to avoid a single fetch taking the rest of the page down with it.

I noticed this while using `testnet` queries with the `fetchTopHolders`, where the fetch often fails. If this does happen- we merely want to log it and return an empty value.

Here I add default values to the following:

- [x] fetching chat events now returns `[]` on failure
- [x] fetching swap events now returns `[]` on failure
- [x] fetching APT price now returns `undefined` on failure
- [x] fetching top holders now returns `[]` on failure
- [x] They all log a warning on failure now

NOTE: The only one this isn't possible for (currently) is the `marketView`, since it utilizes state from on-chain (current `PeriodicStateTrackers`, not in the database), so that one is left alone.

# Testing

I'll test it locally with a failed fetch.

# Checklist

### DIFF
--- a/src/typescript/frontend/src/app/market/[market]/page.tsx
+++ b/src/typescript/frontend/src/app/market/[market]/page.tsx
@@ -30,7 +30,7 @@ interface EmojicoinPageProps {
 const EVENTS_ON_PAGE_LOAD = 25;
 
 const logAndReturnValue = <T extends [] | undefined | null>(dataType: string, onFailure: T) => {
-  console.warn(`Failed to fetch ${dataType}`);
+  console.warn(`[WARNING]: Failed to fetch ${dataType}.`);
   return onFailure;
 };
 
--- a/src/typescript/frontend/src/app/market/[market]/page.tsx
+++ b/src/typescript/frontend/src/app/market/[market]/page.tsx
@@ -29,6 +29,11 @@ interface EmojicoinPageProps {
 
 const EVENTS_ON_PAGE_LOAD = 25;
 
+const logAndReturnValue = <T extends [] | undefined | null>(dataType: string, onFailure: T) => {
+  console.warn(`Failed to fetch ${dataType}`);
+  return onFailure;
+};
+
 export async function generateMetadata({ params }: EmojicoinPageProps): Promise<Metadata> {
   const { market: marketSlug } = params;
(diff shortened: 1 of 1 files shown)


src/typescript/frontend/src/app/market/[market]/page.tsx
```diff
@@ -29,6 +29,11 @@ interface EmojicoinPageProps {
 
 const EVENTS_ON_PAGE_LOAD = 25;
 
+const logAndReturnValue = <T extends [] | undefined | null>(dataType: string, onFailure: T) => {
+  console.warn(`[WARNING]: Failed to fetch ${dataType}.`);
+  return onFailure;
+};
+
```
Comment (CRBl69) [inline src/typescript/frontend/src/app/market/[market]/page.tsx:36]: This is a nice initiative.

I think having a more "top level"/"well-defined" way of handling logs would be nice. Maybe using a logging library ? I often find myself adding `console.log`s to debug stuff. Having some debug logs that can easly be turned on and off would be *sweet*.

I think this is good enough for now though.

src/typescript/frontend/src/app/market/[market]/page.tsx
```diff
@@ -80,14 +85,18 @@ const EmojicoinPage = async (params: EmojicoinPageProps) => {
...
+      ),
       wrappedCachedContractMarketView(marketAddress),
-      getAptPrice(),
-      fetchCachedTopHolders(marketAddress),
+      getAptPrice().catch(() => logAndReturnValue("APT price", undefined)),
+      fetchCachedTopHolders(marketAddress).catch(() => logAndReturnValue("top holders", [])),
       fetchMelee({})
         .then((res) => (res ? res.melee : null))
-        .catch(() => null),
+        .catch(() => logAndReturnValue("arena melee data", null)),
```
Comment (CRBl69) [inline src/typescript/frontend/src/app/market/[market]/page.tsx:99]: I'm wondering if it wouldn't make more sense to have these variables type be `T | ErrorType` or `T | null`.

Ideally, we would have something like `T | Error | Loading` and the inner components that take these values would handle that ?

I feel like by passing an empty array here, we lose all information related to the error.

Anyway, this is an opinion, and I know that doing that would take ages of refactoring.

Your code fixes the issue we're encountering, so I'm approving.

PR #692, opened 2025-03-11
[Do not merge]: Put `arena` in a PR to get an easy preview link

xbtmatt:
# **_Do not merge!_**

But it's not a big deal if you do.

The changes made in this PR are random and don't do anything significant.

I just needed to make some random trivial change to be able to make a PR.

This PR exists to provide a preview link for `arena`.

### DIFF
--- a/src/typescript/frontend/src/app/api/chats/route.ts
+++ b/src/typescript/frontend/src/app/api/chats/route.ts
@@ -1,4 +1,5 @@
 import { stringifyJSON } from "utils";
+import { parseSearchParams } from "utils/url-utils";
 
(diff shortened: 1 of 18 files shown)


src/typescript/frontend/next.config.mjs
```diff
@@ -1,11 +1,6 @@
 // cspell:word dexscreener
 
 // @ts-check
-import analyzer from "@next/bundle-analyzer";
```
alnoki, line 4: Are we sure we want to do this?

Like really really sure?

I mean

I guess it's not a big deal if we do

alnoki (changes requested):
I can not responsibly approve this

PR #693, opened 2025-03-11
fix(ECO-3019): Fix date parsing in `parseJSON`

xbtmatt:
# Description

The regex for a date-like string in our JSON parsing function was slightly incorrect due to a typo. This fixes it.

# Testing

Added tests

# Checklist


CRBl69 (approved):
LGTM.

Gotta love the code working once every ten days

PR #699, opened 2025-03-12
feat(ECO-3028): Add support for granular arena period subscriptions in the broker [3.3/4]

xbtmatt:
# Description

Instead of just a boolean flag, enforce that the broker only accepts a single period type for a subscription to arena candlestick events. Otherwise, the broker sends out 6+ messages to every client on each arena swap.

This way, it only sends out one candlestick per client subscription, based on the period they subscribe to.

- [x] Make the broker accept a single period type for a subscription
- [x] Only send an addition/removal a single time, because they are now "consumed" by the broker each time (added/removed from the hash set of arena candlestick periods) 
- [x] Update the SDK types to use the new type
- [x] Have the home page by default subscribe to "60" aka 1h candlesticks for arena when on the arena page (updated subscriptions from the charting library to come in the charting PR)
- [x] Consolidate the default "resolution" to an exported constant variable and use it
- [x] Update misc TradingView config types to decrease the diff size on the next PR
- [x] Update README.md for the broker and update examples in the typescript SDK README since it explains how to use it (and it's now outdated)

# Testing

- [x] Added simple json serialization unit tests to ensure it works as expected

# Checklist

### DIFF
--- a/src/rust/broker/README.md
+++ b/src/rust/broker/README.md
@@ -37,37 +37,97 @@ Note that they're typed *exactly* as shown below. If you try to send the JSON
 payload over multiple lines through `websocat`, it will error out and parse
 the JSON payload incorrectly.
 
-### Arena events
+### Basic subscriptions
 
-Arena events don't follow the same rules as the other subscriptions.
+The fields `markets`, `event_types`, and `arena` all work by overwriting the
+current subscription state each time a message is received by the broker from
+the client.
 
-To subscribe to arena events, simply pass `{ "arena": true }`. The default value
-is `false`; i.e., no subscription to arena events.
+There is currently no granular control over arena event types, the only
+way to subscribe to them is by simply passing `{ "arena": true }`. The default
+value is `false`; i.e., no subscription to arena events.
 
-To subscribe to arena candlesticks, pass `{ "arena_candlesticks": true }`. The
-default value is `false`.
-
-```json5
-// Subscribe to every single event type.
-{ "arena": true, "arena_candlesticks": true, "markets": [], "event_types": [] }
-
-// Subscribe to all non-arena event types.
-// All of the JSON objects below are equivalent.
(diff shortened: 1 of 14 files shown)


src/typescript/frontend/src/components/charts/const.ts
```diff
@@ -38,23 +39,37 @@ export const ResolutionStringToPeriod: { [key: string]: Period } = {
...
-    "use_localstorage_for_settings" as ChartingLibraryFeatureset,
-    "left_toolbar" as ChartingLibraryFeatureset,
-    "control_bar" as ChartingLibraryFeatureset,
-    "study_templates" as ChartingLibraryFeatureset,
-    "snapshot_trading_drawings" as ChartingLibraryFeatureset,
+    "use_localstorage_for_settings",
+    "left_toolbar",
+    "control_bar",
+    "study_templates",
+    "snapshot_trading_drawings",
```
xbtmatt, line 72: The last 5 lines are just removing unnecessary types (using `as T` can be dangerous as it skips type checking on lots of things)

Then the `enabled_features` has to have `seconds_resolution` to get proper resolutions for seconds-based resolutions on the chart

src/rust/broker/README.md
```diff
@@ -37,37 +37,95 @@ Note that they're typed *exactly* as shown below. If you try to send the JSON
...
+
+// Subscribe to all event types for markets 1 and 2 and effectively unsubscribe 
+// from arena events.
+{ "markets": [1, 2] }
+```
 
+### Arena candlesticks
+
+Arena candlesticks don't follow the same rules as the other subscriptions. To
+subccribe to arena candlesticks, you must pass the specific arena period
```
xbtmatt, line 83: fix (you have a few local changes you didn't push yet)

CRBl69 (approved):

alnoki (approved):

PR #702, opened 2025-03-13
chore(ECO-3043): Misc cleanup, remove unused functions, simplify/fix db model type guards [3.1/4]

xbtmatt:
# Description

- [x] Remove dead/unused code
- [x] Simplify and fix db model type guards
- [x] Update `ArenaClient.tsx` code and remove unused paths/data (it will be added back later- currently isn't used at all)

# Testing

- [x] Visit a market page and ensure candlesticks still works (it should, I tested it)

# Checklist

### DIFF
--- a/src/typescript/frontend/src/app/arena/page.tsx
+++ b/src/typescript/frontend/src/app/arena/page.tsx
@@ -1,15 +1,18 @@
-import { fetchArenaInfo, fetchMarketStateByAddress } from "@/queries/arena";
+import { fetchArenaInfo } from "@/queries/arena";
 import { ArenaClient } from "components/pages/arena/ArenaClient";
-// import { emoji } from "utils";
-// import type { SymbolEmoji } from "@sdk/emoji_data";
-import type { PeriodicStateEventModel } from "@sdk/indexer-v2/types";
 import { redirect } from "next/navigation";
-import { parseJSON } from "utils";
-import { Period } from "@sdk/const";
-import { getCandlesticksRoute } from "app/api/candlesticks/utils";
-import { type AnyNumberString } from "@sdk-types";
+import { ROUTES } from "router/routes";
+import { fetchSpecificMarkets } from "@sdk/indexer-v2";
 
 export const revalidate = 2;
+
+const logAndReturnValue = <T extends [] | undefined | null | Record<string, unknown>>(
+  dataType: string,
+  onFailure: T
+) => {
+  console.warn(`[WARNING]: Failed to fetch ${dataType}.`);
+  return onFailure;
+};
 
 export default async function Arena() {
   let arenaInfo: Awaited<ReturnType<typeof fetchArenaInfo>> = null;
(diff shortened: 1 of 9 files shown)


src/typescript/frontend/src/app/arena/page.tsx
```diff
@@ -1,62 +1,57 @@
...
 
   return (
     <ArenaClient
       arenaInfo={arenaInfo}
       market0={market0!}
       market1={market1!}
-      candlesticksMarket0={candlesticksMarket0}
-      candlesticksMarket1={candlesticksMarket1}
+      candlesticksMarket0={[]}
+      candlesticksMarket1={[]}
```
xbtmatt, line 54: Empty because they are populated in the follow-up PR, and I wanted to make it more clear what's changing from PR to PR.

src/typescript/frontend/src/app/arena/page.tsx
```diff
@@ -1,62 +1,57 @@
...
+      market1: res.find((v) => v.market.marketAddress === arenaInfo.emojicoin1MarketAddress),
+    }))
+    .catch(() =>
+      logAndReturnValue("arena market0 and market1", { market0: undefined, market1: undefined })
+    );
+
+  if (!market0 || !market1) {
+    console.warn("Couldn't fetch market state for one of the arena markets.");
+    redirect(ROUTES.home);
+  }
```
xbtmatt, line 46: Just consolidating this into a single fetch/function with the new query we have, and as mentioned below, removing the old candlestick fetches (See comment below)

src/typescript/frontend/src/components/charts/const.ts
```diff
@@ -19,7 +17,7 @@ export const TV_CHARTING_LIBRARY_RESOLUTIONS = [
   "15",
   "30",
   "60",
-  "4H",
+  "240",
```
xbtmatt, line 20: Apparently either work (I tested it several times), so I just updated it to be consistent with the other array.

src/typescript/frontend/src/components/pages/arena/ArenaClient.tsx
```diff
@@ -101,7 +101,7 @@ export const ArenaClient = (props: ArenaProps) => {
   const [position, setPosition] = useState<ArenaPositionModel | undefined | null>(null);
   const [history, setHistory] = useState<ArenaLeaderboardHistoryWithArenaInfoModel[]>([]);
 
-  useReliableSubscribe({ eventTypes: ["Swap"], arena: true, arenaCandlesticks: true });
+  useReliableSubscribe({ eventTypes: ["Chat"], arena: true });
```
xbtmatt, line 104: No longer need to subscribe to swap events bc we subscribe to arena candlesticks now. And the charting library handles it, so we don't need to even specify it here.

We do, however, need to subscribe to chat events for the chat box stuff to work, so I added it in here.

src/typescript/frontend/src/lib/store/event/candlestick-bars.ts
```diff
@@ -63,9 +63,6 @@ export const toBar = (event: PeriodicStateEventModel): Bar => ({
   volume: toNominal(event.periodicState.volumeQuote),
 });
 
-export const toBars = (events: PeriodicStateEventModel | PeriodicStateEventModel[]) =>
-  Array.isArray(events) ? events.map(toBar) : toBar(events);
-
```
xbtmatt, line 68: We didn't use this

CRBl69 (approved):

PR #619, opened 2025-02-16
feat(ECO-2850): Switch to TradingView charts for arena [4/4]

xbtmatt:
# Description

Update to use the TradingView charts for arena stuff.
This mostly updates the state store and the tradingview chart styling stuff for arena candlesticks.

# Testing

Lots of manual testing.

# Checklist

- [x] Did you update relevant documentation?
- [x] Did you add tests to cover new code or a fixed issue?
- [x] Did you update the changelog?
- [x] Did you check all checkboxes from the linked Linear task?

PR #703, opened 2025-03-14
chore(ECO-3045): Add the new `/api/arena/candlesticks` route; properly implement `Period | ArenaPeriod` types [3.2/4]

xbtmatt:
# Description

Lots of the changes for the new arena candlesticks are simply because the new ArenaPeriod type wasn't properly unioned with Period (on my advice).

I've fixed them to work everywhere applicable so that the utility functions for `Period` types work with both types, but also still requiring that they specifically should only be the base `Period` (non `Arena Period`) type where applicable.

# Testing

Lots of mostly type changes here, but I've been using it in #619 

# Checklist

### DIFF
--- a/src/typescript/frontend/src/app/api/arena/candlesticks/fetch.ts
+++ b/src/typescript/frontend/src/app/api/arena/candlesticks/fetch.ts
@@ -1,0 +1,140 @@
+// cspell:word timespan
+
+import { parseJSON, stringifyJSON } from "utils";
+import { type ArenaCandlesticksSearchParams } from "./search-params-schema";
+import {
+  getCachedLatestProcessedEmojicoinTimestamp,
+  getPeriodDurationSeconds,
+  HISTORICAL_CACHE_DURATION,
+  indexToParcelEndDate,
+  indexToParcelStartDate,
+  jsonStrAppend,
+  NORMAL_CACHE_DURATION,
+  PARCEL_SIZE,
+  toIndex,
+} from "app/api/candlesticks/utils";
+import {
+  type ArenaCandlestickModel,
+  fetchArenaCandlesticksSince,
+  fetchArenaInfoByMeleeID,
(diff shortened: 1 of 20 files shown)


src/typescript/frontend/src/app/api/arena/candlesticks/fetch.ts
```diff
@@ -0,0 +1,140 @@
...
+      data = jsonStrAppend(data, res.data);
+    }
+    if (endDate < meleeStartPeriodBoundary) {
+      break;
+    }
+    i += 1;
+  }
+
+  return data;
+};
```
xbtmatt, line 140: I hand copied all of this from the existing implementation. The main change was the conversion of the times, since one is in microseconds and the other is in milliseconds. Aside from that, the logic is almost exactly the same. You can probably do a simple diff on the files and see the main differences.

CRBl69: Not sure how the original implementation was implemented, but wouldn't it categorize that as a "current" candlestick and only cache it for a short amount of time ? Anyway, that should be fixed by the new parcel PR when it merges.

CRBl69 (approved):

src/typescript/frontend/src/app/api/arena/candlesticks/fetch.ts
xbtmatt: I think you are right- as long as the `indexToParcelEndDate(localIndex, period);` isn't too far out. 

But I think it could possibly be far out? I wasn't sure:

```typescript
export const indexToParcelEndDate = (index: number, period: AnyPeriod): Date =>
  new Date((PARCEL_SIZE * ((index + 1) * periodEnumToRawDuration(period))) / 1000);
```

alnoki (approved):

PR #704, opened 2025-03-14
test(ECO-3046): Parallelize frontend and SDK tests in CI

xbtmatt:
# Description

- [x] Also update `package.json` commands to be more accurate/clear
- [x] Also remove unused commands and commands that don't work

# Testing

They should all run in parallel...and succeed, hopefully!

# Checklist


alnoki (approved):

.github/workflows/tests.yaml
```diff
@@ -28,14 +28,31 @@ jobs:
...
+        - job-name: 'SDK: [e2e, base, general]'
+          command: 'pnpm run test:sdk:e2e:base-general'
+        - job-name: 'SDK: [e2e, base, websockets]'
+          command: 'pnpm run test:sdk:e2e:base-ws'
+        - job-name: 'SDK: [e2e, arena, general]'
+          command: 'pnpm run test:sdk:e2e:arena-general'
+        - job-name: 'SDK: [e2e, arena, websockets]'
+          command: 'pnpm run test:sdk:e2e:arena-ws'
+        - job-name: 'SDK: [e2e, arena, candlesticks]'
+          command: 'pnpm run test:sdk:e2e:arena-candlesticks'
```
alnoki, line 54: How about cleaning this up a little? Just for more tidy printouts in checks tab:

Screenshot 2025-03-14 at 11 56 38

alnoki: In fact I think you could even take out `job-name` and just remove the `name` field entirely, since you'll see the run command

Also, note that each `command` has `pnpm run test:`, so it might be worth using `command-suffix` so that the only think that shows up in the printout is, for example `frontend: unit`

Somethin like       `run: 'pnpm run test:${{ matrix.command-suffix }}'

PR #709, opened 2025-03-17
feat(ECO-3025): Add emoji color picker

0xForkh:
I included a page for testing the color picker at /test/emojicolor


xbtmatt (commented):
this is great, thanks man, really nice to have this!

Any idea why the colors for some of these seem off? I'm fairly bad with colors (pretty colorblind), but I think these are wrong haha:

image


I'm assuming it's doing some weird rendering where it doesn't use the proper emoji font and the colors are misleading? 

To be clear, these are the emojis I'm using:

🎇🎆

Ah okay yes it immediately makes sense actually now that I paste it in here- it looks like it's resolving the emojis to their non-noto-emoji variants. I think you need to make sure the canvas uses the Noto emoji font somehow

Ah yes even more obvious with the flags:

image

xbtmatt (commented):
I should've mentioned this earlier, but I figured you'd have to do a bespoke implementation anyway...

Here's the way `emoji-mart` checks it: https://github.com/econia-labs/emoji-mart/blob/2db39ada7b8926deb54f4b116dfd95d69915a113/packages/emoji-mart/src/helpers/native-support.ts#L45

But I also think they're running into the same issue, anyway

xbtmatt (commented):
We discussed most of it on call already but main thing here is

1. Making it easy to extend with new emojis
2. Add a simple README.md on how to run the script to reproduce the data & extract the relevant colors (even if it's just running something in the browser, copy/paste or right click -> copy obj to clipboard, etc)
3. You can generally minify the output of the colors in the *.ts file and we can handle it at runtime since it's fairly small in terms of runtime memory (probably like ~4,000 elements?). I think all you really need is the emoji and some associated value for the color. `key: value` or  even`[emoji, r, g, b[]` would be fine imo

xbtmatt (approved):
Looks good, just remove the `#` from the colors and add it at runtime. It just sends an extra unnecessary 4-5kb

Also, bit of a nit, but do you mind removing `noto` and `emojicolor` from cspell? Those might actually be typos in most cases so I'd rather not add them globally

cfg/cspell-dictionary.txt
```diff
@@ -80,6 +81,7 @@ mybuilder
 nazar
 nextjs
 nohup
+noto
```
xbtmatt, line 84: Remove this one for sure, v likely to be an actual typo in most cases

src/typescript/frontend/src/lib/utils/emojiColors/emoji-color-data.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+};
+
+const getSingleEmojiColor = async (emoji: SymbolEmoji, dataSource: Record<string, string>) => {
+  if (!(emoji in dataSource)) {
+    console.error(`Emoji ${emoji} not found in color data. Returning default color`);
+    return "#000000";
+  }
+
+  return dataSource[emoji];
```
xbtmatt, line 43:
```suggestion
  return `#${dataSource[emoji]}`;
```

xbtmatt (commented):
Could you also add a couple simple unit tests that check that the # of keys in each record matches the exact number of keys in `SYMBOL_EMOJIS` ?

src/typescript/frontend/src/app/dev/color-generator/page.tsx
```diff
@@ -0,0 +1,9 @@
+import { EmojiColorGenerator } from "lib/utils/emojiColors/emoji-color-generator";
+import { redirect } from "next/navigation";
+
+export default function EmojiColorPage() {
+  if (process.env.NODE_ENV === "production") {
+    redirect("/404");
+  }
```
xbtmatt, line 7:
```suggestion
import { EmojiColorGenerator } from "lib/utils/emojiColors/emoji-color-generator";
import { redirect } from "next/navigation";

import { VERCEL_TARGET_ENV } from "@/sdk/const";

export default function EmojiColorPage() {
  if (VERCEL_TARGET_ENV === "production") {
    redirect("/404");
  }
```

src/typescript/frontend/src/lib/utils/emojiColors/emoji-color-data.ts
```diff
@@ -0,0 +1,53 @@
...
+  return `#${dataSource[emoji]}`;
+};
+
+export const getEmojiColor = async (emoji: SymbolEmoji[], userAgent?: string) => {
+  if (!userAgent) userAgent = window.navigator.userAgent;
+  const { isIOS, isMacOs } = getBooleanUserAgentSelectors(userAgent);
+  const dataSource = await getColorModule(isIOS || isMacOs || false);
+  const colors = await Promise.all(emoji.map((e) => getSingleEmojiColor(e, dataSource)));
+  return colors;
+};
```
xbtmatt, line 53:
```suggestion
export const getEmojiColor = async (emoji: SymbolEmoji[], userAgentArg?: string) => {
  const userAgent = userAgentArg || window?.navigator.userAgent || "";
  const { isIOS, isMacOs } = getBooleanUserAgentSelectors(userAgent);
  const dataSource = await getColorModule(isIOS || isMacOs || false);
  const colors = await Promise.all(emoji.map((e) => getSingleEmojiColor(e, dataSource)));
  return colors;
};
```

It's not really a big deal here, but I'd rather not reassign variables on principle 😁

alnoki (approved):
Approving for line count override

xbtmatt (approved):

PR #713, opened 2025-03-18
feat(ECO-3057): Add/fix various page metadata titles; lazily resolve wallet page slug to ANS name

xbtmatt:
# Description

Some page metadata title values are missing or stylistically inconsistent (use caps)

I've made them all consistent now. I also added the functionality to resolve wallet page ans name inputs to their corresponding addresses.

- [x] Metadata `title`s are now consistent stylistically (lowercase)
- [x] Inputs to the `/wallet` page resolve to an ANS name lazily if it's a valid ANS name
- [x] Otherwise it fetches the ANS name if it's an address
- [x] Otherwise it passes the raw input (but no longer tries to fetch their holdings if it's invalid)

This also means we can display things like matt's wallet | emojicoin.fun in the page metadata title!

https://github.com/user-attachments/assets/4c067d8a-b45e-4674-aa5b-60380e44c2b5

image

image

# Testing

Tested it locally a bunch.

# Checklist


src/typescript/frontend/src/app/verify/page.tsx
```diff
@@ -4,12 +4,18 @@ import {
...
+import { type Metadata } from "next";
 import { cookies } from "next/headers";
 import { redirect } from "next/navigation";
 import { ROUTES } from "router/routes";
 
 export const dynamic = "force-dynamic";
 
+export const metadata: Metadata = {
+  title: "verify",
+  description: "verify your wallet to access all emojicoin.fun features!",
```
CRBl69, line 16: Let's capitalize this (only the description) so it's consistent.

src/typescript/frontend/src/app/wallet/[address]/page.tsx
```diff
@@ -2,18 +2,31 @@ import { WalletClientPage } from "components/pages/wallet/WalletClientPage";
...
-export default async function WalletPage({ params }: { params: { address: string } }) {
+export async function generateMetadata({ params }: Props): Promise<Metadata> {
+  const input = params.address;
+
+  const { name, address } = await resolveOwnerNameCached(input);
+  const owner = name ?? (address ? customTruncateAddress(address) : input);
+
+  return {
+    title: `${owner}'s wallet`,
+    description: `👀 peep what's inside ${owner}'s wallet— are they stacking, trading, or just vibing?`,
```
Comment (CRBl69) [inline src/typescript/frontend/src/app/wallet/[address]/page.tsx:19]: Let's capitalize this (only the description) so it's consistent.

src/typescript/frontend/src/app/wallet/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+  | {
+      // Input resolves to a valid address but has no corresponding ANS name.
+      address: `0x${string}` | `0x${string}...`;
+      name?: undefined;
+    }
+  | {
+      // Input is invalid. It's not an address or an Aptos name.
+      address?: undefined;
+      name?: undefined;
```
CRBl69, line 24: `?` and `undefined` ?

Doesn't this mean `undefined` or `undefined` ?

I know the TS checker treats this ***slightly*** different, but I don't know if it's relevant here.

CRBl69: Same for the `name?` just above.

CRBl69 (changes requested):
Love the page descriptions.

Small consistency changes needed, and then it's good to merge.

src/typescript/frontend/src/app/verify/page.tsx
xbtmatt, line 16:
```suggestion
  description: "Verify your wallet to access all emojicoin.fun features!",
```

src/typescript/frontend/src/app/cult/page.tsx
```diff
@@ -4,8 +4,8 @@ import CultClientPage from "components/pages/cult/CultClientPage";
 export const dynamic = "force-static";
 
 export const metadata: Metadata = {
-  title: "Explore the cult",
-  description: `Explore the emojicoin cult`,
+  title: "cult",
+  description: `The emojicoin cult awaits...🩸`,
```
xbtmatt, line 8:
```suggestion
  description: `The emojicoin cult awaits...👹`,
```

src/typescript/frontend/src/app/wallet/[address]/page.tsx
Comment (xbtmatt) [inline src/typescript/frontend/src/app/wallet/[address]/page.tsx:19]:
```suggestion
    description: `👀 Peep what's inside ${owner}'s wallet— are they stacking, trading, or just vibing?`,
```

src/typescript/frontend/src/app/wallet/utils.ts
xbtmatt: Actually you have a good point, I think I meant to do the opposite of this. But the difference is kind of complicated. Basically I'm doing two things here, only one of which I want to do (so I'll change it)

#### The thing I want to do
I'm creating explicit discriminated types so that typescript enforces me to return a valid combination of the two. For example, if I tried to do `return { address, name }` where `address` is potentially undefined but `name` is `ValidAptosName`, the compiler wouldn't let me return that, as it would say that's an invalid return type. In this case in particular, I do that because there's no case where the name is valid but the address is not (because an ANS name will always have a corresponding valid address)

#### The thing I didn't mean to do
When you return `field?: number` vs `field: number | undefined`, the primary difference is that in the former, `field` can be missing entirely from the object return type. So `{}` would be valid for the former, but not for the latter. The latter would explicitly require that you list the field in the return type: `{ field }` even if field itself is actually just undefined.

Sometimes this is undesirable if a code path doesn't ever declare `field`, because then ts will say "`field` is not a valid variable" or whatever it says when you try to use something uninstantiated

But in terms of how the function return type is *used* it's generally similar- although leaving `field` out entirely means you could get a `ReferenceError` if `field` is an object, and you try to unpack a field inside of *it*:

```typescript
type response = {
  field?: undefined |
    {
      inner?: number
    }
 };

// no error:
const { field } = response; 
// typescript throws a `ReferenceError` here:
const { inner } = field;
```

But the primary reason for specifying `field?: undefined` in most cases is actually so you can use function overloads properly (for the input arg type or the return type for example)- if you don't use `field?: T` but use `field: T | undefined`, then the user explicitly has to pass `field: undefined` in order for typescript to resolve to the proper overloaded type

Anyway, I addressed it in 04af987a, haha. I just made it required to return `undefined`, mostly for point 1

CRBl69 (approved):

PR #716, opened 2025-03-19
chore(ECO-3076): Kill alpha stack, update indexer binaries

alnoki:
# Description

1. Kill the `emoji-alpha` stack
2. Update binaries across indexer contexts

# Testing

CI, once processor finishes building and pushing to Docker Hub: https://github.com/econia-labs/aptos-indexer-processors/actions/runs/13956329437

# Checklist


xbtmatt (commented):
We don't need the `*-fallback` or `-*production` yaml files for these right? Just wondering if there's a reason you add them

alnoki:
> We don't need the `*-fallback` or `-*production` yaml files for these right? Just wondering if there's a reason you add them

By killing both of them, there is one less PR required for the production release: just a single PR that revives both stacks

Then it is possible to merge two sequential commits over to production for the cold upgrade

xbtmatt (approved):
Ah yes I remember why you do it now- so we can pull up to a certain point and not have to make a separate PR for killing those stacks as well.

Approving since draft will block merge for now

PR #717, opened 2025-03-20
feat(ECO-3074): Add emojicoin arena feature flag; feature gate it where applicable

xbtmatt:
# Description

- [x] Assume arena is off unless there is an environment variable set that says it is on
- [x] Can feature gate it with env vars without anything breaking
- [x] Gate all frontend features wherever applicable
- [x] Added build-time checks to ensure that a `NEXT_PUBLIC_ARENA_MODULE_ADDRESS` has a valid `emojicoin_arena.move` contract deployed if the arena feature flag `NEXT_PUBLIC_ARENA_ENABLED` is set to `"true"`

# Testing

Manual testing, mainly just visiting all pages affected with the env var set and not set.

Can also test it with invalid API keys or invalid endpoints to test if things continue to work without melee data.

# Checklist


src/typescript/frontend/src/lib/feature-flags.ts
```diff
@@ -0,0 +1,5 @@
+export const FEATURE_FLAGS = {
+  Arena: process.env.NEXT_PUBLIC_ARENA_ENABLED === "true",
```
CRBl69, line 2: The rate limiting switch was recently implemented to take "1" when enabled. Let's be consistent and do the same here.

CRBl69 (approved):

alnoki (approved):

src/typescript/README.md
```diff
@@ -38,6 +38,16 @@ The integrator address for the contract, aka the address that receives fees.
 
 - **Example**: `"0x33332c9ea4c220e0572b7f83f397164f8171e1c9f681136bb8ab78efa6c43333"`
 
+#### `NEXT_PUBLIC_ARENA_MODULE_ADDRESS`
+
+The contract address of the arena module.
```
alnoki, line 43:
```suggestion
The emojicoin arena package address.
```

src/typescript/frontend/src/lib/feature-flags.ts
CRBl69: It was actually Matt's suggestion to use `1`. There is no JS standard for env vars.

CRBl69: Just checked and we actually have 3 env vars, two using `true` and one using `1`. Let's use `true` here and fix the other one in another PR.

CRBl69: Nevermind, I see matt already did that

src/typescript/README.md
alnoki: Marking resolved, can tidy up later to stop merge block

PR #721, opened 2025-03-20
test(ECO-3082): Move ts `format` and `lint` into a GH Action, add `check` for compilation errors

xbtmatt:
# Description

- [x] Speed up pre-commit by removing `format` and `lint`
- [x] Make it more clear when `format` / `lint` fails
- [x] Add the `pnpm run check` command for static type analysis (currently we rely on `Vercel` build or test suites failing)
- [x] Re-use all the cached pnpm dependencies in a shared cache in the workflow (used an action for it)

~- [ ] (Maybe for another PR?) Add a check for the output of `.next/app-paths-manifest.json` upon build, which we can use instead of our current `ROUTES` var in `routes.ts`~ Will do this in a different PR

# Testing

Built into the PR

# Checklist


.github/workflows/ts-checks.yaml
```diff
@@ -1,27 +1,57 @@
...
-        cache-dependency-path: 'src/typescript/pnpm-lock.yaml'
-        node-version-file: 'src/typescript/.node-version'
-    - name: 'Install dependencies'
-      run: 'pnpm install'
+        submodules: 'false'
+    - uses: './.github/actions/pnpm-install'
+    - run: 'pnpm run check'
+
+  tests:
+    name: 'test suite / (${{ matrix.pretty-suffix }})'
```
xbtmatt, line 48:
```suggestion
    name: 'test / (${{ matrix.pretty-suffix }})'
```

.github/workflows/ts-checks.yaml
```diff
@@ -1,27 +1,57 @@
...
 ---
+name: 'typescript'
+
+'on':
+  merge_group: null
+  pull_request: null
+  push:
+    branches:
+    - 'main'
+    - 'production'
```
alnoki, line 12: This can actually be taken out, since the `pull_request` and `merge_group` triggers will ensure that it runs on a pull request into the branches listed here

alnoki (approved):

.github/workflows/ts-checks.yaml
xbtmatt: Oooh ok thanks, removed it in 2b55e07d

.github/workflows/ts-checks.yaml
```diff
@@ -1,27 +1,53 @@
...
+    working-directory: 'src/typescript'
+
 jobs:
-  typescript-tests:
-    defaults:
-      run:
-        working-directory: 'src/typescript'
-    name: 'typescript / (${{ matrix.pretty-suffix }})'
+  lint:
     runs-on: 'ubuntu-latest'
```
alnoki, line 17:
```suggestion
  lint:
    name: 'TS lint'
    runs-on: 'ubuntu-latest'
```

.github/workflows/ts-checks.yaml
```diff
@@ -1,27 +1,53 @@
...
       with:
-        package_json_file: 'src/typescript/package.json'
-    - name: 'Install node.js'
-      uses: 'actions/setup-node@v4'
+        submodules: 'false'
+    - uses: './.github/actions/pnpm-install'
+    - run: 'pnpm run format:check'
+
+  tsc-check:
+    runs-on: 'ubuntu-latest'
```
alnoki, line 35:
```suggestion
  ts-check:
    name: 'TS check'
    runs-on: 'ubuntu-latest'
```

.github/workflows/ts-checks.yaml
```diff
@@ -1,27 +1,53 @@
...
     - uses: 'actions/checkout@v4'
       with:
         submodules: 'false'
-    - name: 'Install pnpm'
-      uses: 'pnpm/action-setup@v4'
+    - uses: './.github/actions/pnpm-install'
+    - run: 'pnpm run lint'
+
+  format:
+    runs-on: 'ubuntu-latest'
```
alnoki, line 26:
```suggestion
  format:
    name: 'TS Format'
    runs-on: 'ubuntu-latest'
```

.github/workflows/ts-checks.yaml
xbtmatt: this one is actually intentional bc `tsc` is the command, so it's checking that tsc (typescript compile) succeeds

.github/workflows/ts-checks.yaml
xbtmatt: for the rest, it's already prefixes with `typescript` in the overall workflow. But I also am not familiar with everywhere this might appear so I'll defer to you on what is the clearest! Just wanted to note it will appear as `typescript / TS format`

.github/workflows/ts-checks.yaml
alnoki: Ahh okay, thanks - updated the suggestion

.github/workflows/ts-checks.yaml
alnoki: Ya it's mainly so it is less ambiguous when updating branch settings, since branch settings only use the job name, not the strcat(workflow name, job name)

xbtmatt: yeah I thought about it and you're right it's clearer. Will update all in a sec!

PR #731, opened 2025-03-21
refactor(ECO-3093): Refactor `providers.tsx` and remove unnecessary ContextProviders

xbtmatt:
# Description

There are a lot of context providers that don't need to exist. With `zustand`, you only need to use context providers when you need to employ dependency injection for the initial store.

The only ones we do this with are the user settings store and the emoji picker store, so they remain the same.

- [x] Remove unnecessary context providers
- [x] Added a `layout.tsx` for `arena` which will let us fetch data in the layout as a server component.
- [x] Remove some old code in `providers.tsx` that is no longer used.

# Testing

Using the frontend and ensuring events still appear in the various state stores

# Checklist


src/typescript/frontend/src/context/event-store-context/hooks.tsx
```diff
@@ -35,3 +15,12 @@ export const useUserSettings = <T,>(selector: (store: UserSettingsStore) => T):
 
   return useStore(userSettingsContext, selector);
 };
+
+export const globalNameStore = createNameStore();
+export const globalEventStore = createEventStore();
```
xbtmatt, line 20: Global stores like these mean we could also remove any functions that exist as stable references to getters- like `getMarketMap` and things like that.

Consider removing those functions and simply using the global event store.

src/typescript/frontend/src/app/arena/layout.tsx
```diff
@@ -0,0 +1,3 @@
+export default async function ArenaLayout({ children }: { children: React.ReactNode }) {
+  return <>{children}</>;
+}
```
CRBl69, line 3: What is the point of an empty layout ?

CRBl69 (changes requested):

src/typescript/frontend/src/app/arena/layout.tsx
xbtmatt: It says in the PR description:

>  Added a layout.tsx for arena which will let us fetch data in the layout as a server component.

We can fetch data in `layout.tsx` and pass it directly to a `ContextProvider` or anything that needs to sync with server data immediately in the global app state.

an example is a zustand store- basically anything that should be available on the first render from the server but also be in state as well. Fetching it in layout makes it easier because it's a server component but directly wraps client component providers, so you can do dependency injection with it on first render

CRBl69 (approved):

src/typescript/frontend/src/lib/store/event/event-store.ts
```diff
@@ -251,6 +251,11 @@ export const createEventStore = () => {
     }))
   );
 
+  // Return early to avoid localStorage operations.
```
xbtmatt, line 254:
```suggestion
  // Return early to avoid state mutations based on localStorage.
```

PR #732, opened 2025-03-21
feat(ECO-3094): Add dev mode

CRBl69:
# Description

This PR adds a toggable dev mode that can be enabled on any deployment.


xbtmatt (commented):
This is great :) But I think we should consider limiting it to non-production environments or at least restricting it to special whitelisted addresses or something. Just because some of it can make more queries/connections and we might add later functionality to it

What do you think?

src/typescript/frontend/src/context/DevMode.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return <DevModeContext.Provider value={devMode}>{children}</DevModeContext.Provider>;
+}
+
+export const useDevMode = (): boolean => {
+  const context = useContext(DevModeContext);
+  if (context === null) {
+    throw new Error("useDevMode must be used within a DevModeContext.");
+  }
+  return context;
+};
```
xbtmatt, line 19: You can get rid of the context provider and just directly access it from within the user settings.

```typescript
const devMode = useUserSettings((s) => s.devMode);

// or as a hook:
export const useDevMode = (() => useUserSettings((s) => s.devMode));
```

But it's so simple I don't think it needs its own hook, rather not add a file for this since it's so simple

src/typescript/frontend/src/context/ConnectToWebSockets.tsx
```diff
@@ -3,14 +3,17 @@ import { motion } from "framer-motion";
 import { emoji } from "utils";
 import { Emoji } from "utils/emoji";
 import { hexToRgba } from "utils/hex-to-rgba";
+import { useDevMode } from "./DevMode";
 
 export const ConnectToWebSockets = () => {
   const connected = useEventStore((s) => s.connected);
   const received = useEventStore((s) => s.received);
 
+  const devMode = useDevMode();
```
xbtmatt, line 12:
```suggestion
  const devMode = useUserSettings((s) => s.devMode);
```

xbtmatt (commented):
Just remove the context provider and pull it directly from UserSettingsStore, other than that looks good! Let's get this in

src/typescript/frontend/src/context/DevMode.tsx
CRBl69: addressed in 92e8d7169d3e3c0481096fb6b057aeec4f6e7e26

PR #733, opened 2025-03-21
fix(ECO-3095): Fix ANS regex

CRBl69:
# Description

This PR fixes the ANS regex.

It was missing the suffix (e.g. `.apt`/`.petra`).

### DIFF
--- a/src/typescript/frontend/src/app/wallet/[address]/error.tsx
+++ b/src/typescript/frontend/src/app/wallet/[address]/error.tsx
@@ -1,0 +1,15 @@
+"use client";
+
+import { emoji } from "utils";
+import { Emoji } from "utils/emoji";
+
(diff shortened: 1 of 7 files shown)


xbtmatt (commented):
This was mostly just supposed to match the rules specified in the config: 0x867ed1f6bf916171b1de3ee92849b8978b7d1b9e0a8cc982a3d19d535dfd9c0c::v2_1_config::Config

I'm not sure how the subdomains work exactly- I think if you wanted to extend the regex you could but this regex was strictly just for the base/primary(?) name

xbtmatt (commented):
Thx for catching this- I added proper validation in case we ever re-use it and just to make it more robust. In #734 

Requested to merge it into here and then we can merge this

xbtmatt (approved):

PR #734, opened 2025-03-21
refactor(ECO-3098): Validate Aptos name inputs properly and display u…

xbtmatt:
…ser not found page upon error

# Description

I didn't really understand the subdomain/domain parts of the ANS name service before, but the TypeScript SDK has internal validation functions already.

Unfortunately, it doesn't export those functions anywhere, and the only other alternative to skipping ANS name fetching if it's an invalid name is with a `try/catch` block, which is bad for maintainability and readability.

To fix this, I ported over [this logic](https://github.com/aptos-labs/aptos-ts-sdk/blob/db88e984528a545a240bad876d9ea832fd0c8348/src/internal/ans.ts#L26) and put it into a zod schema:

```typescript
export const VALIDATION_RULES_DESCRIPTION = [
  "A name must be between 3 and 63 characters long,",
  "and can only contain lowercase a-z, 0-9, and hyphens.",
  "A name may not start or end with a hyphen.",
].join(" ");

/**
 * Validate if a given fragment is a valid ANS segment.
 * This function checks the length and character constraints of the fragment to ensure it meets the ANS standards.
 *
 * @param fragment - A fragment of a name, either the domain or subdomain.
 * @returns A boolean indicating if the fragment is a valid fragment.
 * @group Implementation
 */
export function isValidANSSegment(fragment: string): boolean {
  if (!fragment) return false;
  if (fragment.length < 3) return false;
  if (fragment.length > 63) return false;
  // only lowercase a-z and 0-9 are allowed, along with -. a domain may not start or end with a hyphen
  if (!/^[a-z\d][a-z\d-]{1,61}[a-z\d]$/.test(fragment)) return false;
  return true;
}

/**
 * Checks if an ANS name is valid or not.
 *
 * @param name - A string of the domain name, which can include or exclude the .apt suffix.
 * @group Implementation
 */
export function isValidANSName(name: string): { domainName: string; subdomainName?: string } {
  const [first, second, ...rest] = name.replace(/\.apt$/, "").split(".");

  if (rest.length > 0) {
    throw new Error(`${name} is invalid. A name can only have two parts, a domain and a subdomain separated by a "."`);
  }

  if (!isValidANSSegment(first)) {
    throw new Error(`${first} is not valid. ${VALIDATION_RULES_DESCRIPTION}`);
  }

  if (second && !isValidANSSegment(second)) {
    throw new Error(`${second} is not valid. ${VALIDATION_RULES_DESCRIPTION}`);
  }

  return {
    domainName: second || first,
    subdomainName: second ? first : undefined,
  };
}
```

I also added an `error.tsx` for `/[wallet]` and return the `UserNotFound` component in it if the address or user are undefined.

# Testing

Testing it manually with various names and subdomains.

Added unit tests for ans name validation (test cases copied from the TS SDK).

# Checklist

### DIFF
--- a/src/typescript/frontend/src/lib/hooks/queries/use-fetch-owner-emojicoin-balances.ts
+++ b/src/typescript/frontend/src/lib/hooks/queries/use-fetch-owner-emojicoin-balances.ts
@@ -72,14 +72,22 @@ const toFullCoinData = async (balances: AssetBalance[]) => {
  * @returns the @see FullCoinData[] and wallet stats total value
  */
 export const useUserEmojicoinBalances = (owner: AccountAddressInput, max?: number) => {
-  const ownerAddress = useMemo(() => toAccountAddressString(owner), [owner]);
+  const ownerAddress = useMemo(
+    () => owner && AccountAddress.isValid({ input: owner }).valid && toAccountAddressString(owner),
+    [owner]
+  );
 
   const { data, isLoading } = useQuery({
     queryKey: ["user-emojicoin-balances", ownerAddress, max],
     queryFn: () =>
-      withResponseError(fetchOwnerEmojicoinBalances({ ownerAddress, max }).then(toFullCoinData)),
+      withResponseError(
+        (ownerAddress
+          ? fetchOwnerEmojicoinBalances({ ownerAddress, max })
+          : Promise.resolve([])
+        ).then(toFullCoinData)
+      ),
     staleTime: STALE_TIME,
-    enabled: AccountAddress.isValid({ input: owner }).valid,
+    enabled: !!owner && AccountAddress.isValid({ input: owner }).valid,
   });
 
   return {
--- a/src/typescript/sdk/src/utils/validation/ans-name.ts
+++ b/src/typescript/sdk/src/utils/validation/ans-name.ts
@@ -31,7 +31,7 @@ const AnsNameSchema = z
   .transform((s) => s.replace(/\.apt$/, ""))
   .refine((nameWithoutSuffix) => {
     const parts = nameWithoutSuffix.split(".");
-    return parts.length > 2 && parts.every((p) => AnsSegmentSchema.safeParse(p).success);
+    return parts.length <= 2 && parts.every((p) => AnsSegmentSchema.safeParse(p).success);
   });
 
 export const isValidAptosName = (name: unknown): name is ValidAptosName =>
--- a/src/typescript/sdk/src/utils/validation/ans-name.ts
+++ b/src/typescript/sdk/src/utils/validation/ans-name.ts
@@ -1,31 +1,38 @@
+import { z } from "zod";
+
 export type ValidAptosName = string & {
   __brand: "ValidAptosName";
 };
 
 /**
- * - Only lowercase characters `a-z`, digits `0-9`, or a hyphen `-`
- * - Doesn't start with a hyphen
- * - Doesn't end with a hyphen
- * - At least 3 characters
- * - At most 63 characters
+ * Schema validation largely borrowed from the ANS name validation logic in the Aptos TS SDK.
+ *
+ * Unfortunately it is not exported to the TS SDK, but the logic is fairly straightforward.
+ *
+ * @see {@link https://github.com/aptos-labs/aptos-ts-sdk/blob/db88e984528a545a240bad876d9ea832fd0c8348/src/internal/ans.ts#L55}
  */
-const APTOS_NAME_REGEX = /^[a-z0-9][a-z0-9-]{1,61}[a-z0-9](\.[a-z0-9]+)?$/;
 
-/**
- * Checks to see that an ANS name is valid. This means we can skip API calls to the Aptos Name
- * Service for names that are invalid.
- *
- * The rules used here are taken from the
- * {@link [ANS contract on mainnet](https://explorer.aptoslabs.com/account/0x867ed1f6bf916171b1de3ee92849b8978b7d1b9e0a8cc982a3d19d535dfd9c0c/modules?network=mainnet)}
- *
- * The rules are written in `v2_1_string_validator`:
- *  - Only allow latin lowercase letters, digits, and hyphens
- *  - Hyphens are not allowed at the beginning or end of a string
- *
- * As well as having the `min_domain_length` and `max_domain_length` set to `3` and `63` specified
- * in the
- * {@link [v2_1_config::Config](https://explorer.aptoslabs.com/account/0x867ed1f6bf916171b1de3ee92849b8978b7d1b9e0a8cc982a3d19d535dfd9c0c/resources?network=mainnet)}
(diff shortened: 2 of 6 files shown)


src/typescript/frontend/src/lib/hooks/queries/use-fetch-owner-emojicoin-balances.ts
```diff
@@ -72,14 +72,22 @@ const toFullCoinData = async (balances: AssetBalance[]) => {
  * @returns the @see FullCoinData[] and wallet stats total value
  */
 export const useUserEmojicoinBalances = (owner: AccountAddressInput, max?: number) => {
-  const ownerAddress = useMemo(() => toAccountAddressString(owner), [owner]);
+  const ownerAddress = useMemo(
+    () => owner && AccountAddress.isValid({ input: owner }).valid && toAccountAddressString(owner),
+    [owner]
+  );
```
xbtmatt, line 78: Also added validation here. We should probably have a `ValidAccountAddress` schema as well. This one is actually pretty bad for input validation because it will return `true` for any input that isn't a string or a hex input/account address already...

for example:

```typescript
// prints true...
console.log(AccountAddress.isValid({ input: undefined as any, strict: true }).valid)
```

src/typescript/sdk/src/utils/validation/ans-name.ts
```diff
@@ -1,31 +1,38 @@
...
+const AnsNameSchema = z
+  .string()
+  .transform((s) => s.replace(/\.apt$/, ""))
+  .refine((nameWithoutSuffix) => {
+    const parts = nameWithoutSuffix.split(".");
+    return parts.length <= 2 && parts.every((p) => AnsSegmentSchema.safeParse(p).success);
+  });
+
+export const isValidAptosName = (name: unknown): name is ValidAptosName =>
+  AnsNameSchema.safeParse(name).success;
```
CRBl69, line 38: Maybe I'm missunderstanding something but it seems like `a` is a valid name in [Aptos Labs' tests](https://github.com/aptos-labs/aptos-names-contracts/blob/001d9998d33bb31e693dc7005fc513deabadf748/core_v2/sources/v2_1_string_validator.move#L132-L137).

xbtmatt: Those are just unit tests for latin letter string parsing; the checks for names are here:

https://github.com/aptos-labs/aptos-names-contracts/blob/001d9998d33bb31e693dc7005fc513deabadf748/core_v2/sources/v2_1_domains.move#L908

CRBl69: Ok thanks. I wasn't able to find the TLD validation though. Is that also at least 3 characters ?

xbtmatt: Yes it looks like it calls the same function for both domains and subdomains:

https://github.com/aptos-labs/aptos-names-contracts/blob/001d9998d33bb31e693dc7005fc513deabadf748/core_v2/sources/v2_1_domains.move#L286

https://github.com/aptos-labs/aptos-names-contracts/blob/001d9998d33bb31e693dc7005fc513deabadf748/core_v2/sources/v2_1_domains.move#L313

Note I mostly just copied this code from the TS-SDK, main thing to check here is that I ported it over properly, unless you're wondering if the TS-SDK code is wrong too

CRBl69: I'm not talking about subdomains.

`subdomain.domain.tld`.

TLD is `.apt` or `.petra`.

Could there be a `.ec` for exapmle ?

xbtmatt: The `.apt` is not really a domain in the way you're thinking (if I'm understanding you correctly), it's more just a suffix that exists on every name.

For example this name is a `petra` name:

https://explorer.aptoslabs.com/account/0x67b5fff16f5c0a1ac9eb6409da1fc98899b75e28902cae6a73b8e4b1047976e1?network=mainnet

xbtmatt: was curious to make sure:

```typescript
import { Aptos, AptosConfig, Network } from "@aptos-labs/ts-sdk";

export const main = async () => {
  const mainnet = new Aptos(
    new AptosConfig({
      network: Network.MAINNET,
    })
  );
  await mainnet.ans.getName({ name: "matt" }).then(console.log);
  await mainnet.ans.getDomainSubdomains({ domain: "matt" }).then(console.log);
  await mainnet.ans.getDomainSubdomains({ domain: "petra" }).then(console.log);
};

main();
```

```typescript
// 1. my primary name
{
  domain: 'matt',
  expiration_timestamp: 1764735392000,
  registered_address: '0xffc117086980d34dc3b5a42cb407ed888f60623f46021f35c2ca522ea13cc961',
  subdomain: '',
  token_standard: 'v2',
  is_primary: true,
  owner_address: '0xffc117086980d34dc3b5a42cb407ed888f60623f46021f35c2ca522ea13cc961',
  subdomain_expiration_policy: null,
  domain_expiration_timestamp: '2025-12-02T20:16:32'
}

// 2. me having no subdomains
[]

// 3. petra subdomains
[
  {
    domain: 'petra',
    expiration_timestamp: 1733587274000,
    registered_address: '0x734b3a5f843c0b67c3eb8005cee99405f0ef525e33b439d3e9e4ae6062d70446',
    subdomain: 'joan1',
    token_standard: 'v2',
    is_primary: true,
    owner_address: '0x734b3a5f843c0b67c3eb8005cee99405f0ef525e33b439d3e9e4ae6062d70446',
    subdomain_expiration_policy: 1,
    domain_expiration_timestamp: '2123-09-19T23:38:43'
  },
  {
    domain: 'petra',
    expiration_timestamp: 1712365840000,
    registered_address: '0xf0d9e13d6fa2941a823353b409a81b6fb2edaab9464d84a7c76a4d8541617d72',
    subdomain: 'chubii',
    token_standard: 'v2',
    is_primary: true,
    owner_address: '0xf0d9e13d6fa2941a823353b409a81b6fb2edaab9464d84a7c76a4d8541617d72',
    subdomain_expiration_policy: 1,
    domain_expiration_timestamp: '2123-09-19T23:38:43'
  },
  {
    domain: 'petra',
    expiration_timestamp: 1713977135000,
    registered_address: '0x802bd95284ecb6ac8b43b4162ed2d3d8e2b3824d59febeaeb97a645ff23a0cc2',
    subdomain: 'munna45',
    token_standard: 'v2',
    is_primary: true,
    owner_address: '0x802bd95284ecb6ac8b43b4162ed2d3d8e2b3824d59febeaeb97a645ff23a0cc2',
    subdomain_expiration_policy: 1,
    domain_expiration_timestamp: '2123-09-19T23:38:43'
  },
  // a bunch more
]
```

CRBl69 (approved):

PR #737, opened 2025-03-23
chore(ECO-3101): Remove dead code

xbtmatt:
# Description

Removes dead code that's unused anywhere. As long as everything still builds, it's clear that it's not actually being used


xbtmatt:
Thought I'd use multiple PRs for this but only ended up using one. Closed in favor of #738

PR #739, opened 2025-03-23
feat(ECO-3024): Add the chat box for arena

0xForkh:
Here is a preview of what the chat looks like when we use the emoji dominant color as background color:

image

IMO doesn't look great. It's also problematic when the emoji from the chat message has the same color as the background. Which will happen a lot since most people spam the emoji they're supporting. 

Not sure how to tweak it. Make it darker?

Here is another preview. As expected, the gradient doesn't look great.

image

Maybe it would be better to have 2 fixed colors?


xbtmatt (approved):
Looks awesome- I think you did implement it but just want to make sure you tested it- does this show new chat events coming in through the websocket connection?

src/typescript/frontend/src/app/api/chats/route.ts
```diff
@@ -1,4 +1,5 @@
 import { stringifyJSON } from "utils";
+import { parseSearchParams } from "utils/url-utils";
```
alnoki, line 2: @xbtmatt feel free to resolve this comment if this is handled (I don't want to block merge if I'm just misinformed)

But how does this work for when there are multiple emojis in the chat box?

You can chat with just more than one emoji in a message, but the examples show just single-emoji messages

xbtmatt: The search params only include the marketID, the emojis in the message are just the `message` column as the full string in the chat row entry- does that answer your question or am I misunderstanding you? 😁

alnoki: So firstly I might have instigated some confusion by using a comment on this line (I just wanted to leave a comment somewhere so that it'd have to be addressed before merge)

More generally my question is about the logic for picking what color to render. I ran the preview and I understand that the picture of the chat box is based on the emoji from the market

But has this been tested on  markets with multiple emojis? For example how would a market of `yellow_emoij, blue_emoji` resolve? To yellow? to blue? to brown?

xbtmatt: No worries I figured that might be the case haha, and right now it essentially uses a gradient for the background across the symbol emojis. But you're right to question how it works for multiple emojis- it looks like it supports two emojis in a symbol but not 3.

There might have been a miscommunication somewhere because I was planning on tweaking it a bit before it gets in (so I appreciate you reviewing it first)- but the plan is to use a similar background for all chats but change the border or some auxiliary aspect of it to mark the chat box as belonging to one "side". And to be clear, that is not what's going on in this current implementation- only because the color generated from the multi-emoji symbols is just being used in the wrong place.

But the foundation for how we pick the colors is there and will essentially be some gradient of the multiple emojis in a symbol, or something to that effect

src/typescript/frontend/src/components/pages/arena/tabs/ChatTab.tsx
```diff
@@ -1,10 +1,127 @@
...
+
+  useEffect(() => {
+    setMode("chat");
+    /* eslint-disable-next-line react-hooks/exhaustive-deps */
+  }, []);
+
+  const { data: emoji0Color } = useEmojiColors(market0.market.symbolEmojis);
+  const { data: emoji1Color } = useEmojiColors(market1.market.symbolEmojis);
+
+  const bgColors = useMemo(() => {
```
xbtmatt, line 78: @alnoki Specifically here is how it creates the gradient- although this needs to support 3 emojis as well

src/typescript/frontend/src/components/pages/arena/tabs/ChatTab.tsx
```diff
@@ -1,10 +1,127 @@
...
+          {sortedChats.map(({ message, shouldAnimateAsInsertion }, index) => (
+            <MessageContainer
+              message={message}
+              key={message.version}
+              index={sortedChats.length - index}
+              shouldAnimateAsInsertion={shouldAnimateAsInsertion}
+              alignLeft={message.marketID === market0.market.marketID}
+              backgroundColor={
+                message.marketID === market0.market.marketID ? bgColors?.color0 : bgColors?.color1
+              }
```
xbtmatt, line 107: and this is not being used properly (I think the props are not being applied correctly- will fix this)

src/typescript/frontend/src/app/api/chats/route.ts
alnoki: > planning on tweaking it a bit before it gets in

Sounds like a plan, resolving to let you address in other comments

src/typescript/frontend/src/components/pages/arena/tabs/ChatTab.tsx
alnoki: Sounds like a plan, please do update that in here before it merges, will yield to you on comment resolution

src/typescript/frontend/src/components/pages/arena/tabs/ChatTab.tsx
alnoki: Sounds like a plan, please do update that in here before it merges, will yield to you on comment resolution

src/typescript/frontend/src/components/pages/arena/tabs/ChatTab.tsx
xbtmatt: Kept it simple for now, removed gradients/bg colors and labeled each message w the arena symbol. Can update later but this is clean imo

src/typescript/frontend/src/components/pages/arena/tabs/ChatTab.tsx
xbtmatt: Kept it simple for now, removed gradients/bg colors and labeled each message w the arena symbol. Can update later but this is clean imo

xbtmatt (approved):

PR #742, opened 2025-03-24
chore(ECO-3109): Add `simple-import-sort` to `sdk/.eslintrc.js` rules [2/6]

xbtmatt:
# Description

It's auto-fixable, so I just had to run `pnpm run lint:fix` and it fixed all of it. The only manual additions were changing `require("server-only")` to `import "server-only";`


xbtmatt:
Should be stacked on top of #741 , will remove this and merge once that merges to main

CRBl69 (commented):
Thank you for doing this, it's so nice !

Is there any real/manual change besides the lock file, the package.json files and the .eslint.rc file ?

CRBl69 (approved):

PR #745, opened 2025-03-24
chore(ECO-3114): Convert all type imports to `import type { A } from "a";` with lint rule [5/6]

xbtmatt:
# Description

Fairly self-explanatory, this is the modern way to use types to facilitate more efficient tree-shaking. It also satisfies the new eslint rule and makes importing types more consistent/auto-fixable


xbtmatt:
Should be stacked on top of ECO-3113, will remove this and merge once that merges to main

CRBl69 (approved):

PR #751, opened 2025-03-25
fix(ECO-3124): Fix My Trade History from showing all trade history fo…

0xForkh:


src/typescript/frontend/src/components/pages/wallet/useSwapEventsQuery.ts
```diff
@@ -25,7 +37,7 @@ export const useSwapEventsQuery = (args: z.input<typeof GetTradesSchema>, disabl
     getNextPageParam: (lastPageResponse, allPageResponses) =>
       lastPageResponse?.length === LIMIT ? allPageResponses.length + 1 : undefined,
     // Disable the query when neither sender nor marketID is provided
-    enabled: !disabled && (!!args.sender || !!args.marketID),
+    enabled: options?.enabled !== false && (!!args.sender || !!args.marketID),
```
xbtmatt, line 40:
```suggestion
    enabled: options?.enabled === true && (!!args.sender || !!args.marketID),
```

Should be this, maybe? Or are you requiring an explicit `enabled === false` to disable it?

If so I'd change it to be named something more explicit- in general though it's not ideal to treat `undefined` similar to a truthy value, would rather fix the names and/or behavior to be more intuitive

xbtmatt (approved):
Any reason not to just use `disabled`? I think it's fine to use it  because `enabled` is the default (right?) so you need to explicitly pass `disabled` to disable it

Anyway, not a huge deal, just write a doc comment for it explaining the behavior whatever you decide to do 😁

PR #752, opened 2025-03-25
fix(ECO-3127): Only regenerate background emojis on a step interval to avoid excessive re-renders

xbtmatt:
# Description

See the videos attached below for the difference and explanation (it's fairly obvious):

https://github.com/user-attachments/assets/7babe320-6350-43d5-b51f-3a02e57f9bb3

### DIFF
--- a/src/typescript/frontend/src/components/misc/background-emojis/use-window-size-with-step.ts
+++ b/src/typescript/frontend/src/components/misc/background-emojis/use-window-size-with-step.ts
@@ -1,52 +1,23 @@
 import { useMemo } from "react";
 import useWindowSize from "react-use/lib/useWindowSize";
 
-/**
- * Rounds a numeric value down to the nearest multiple of the specified step size.
- * @param value - The numeric value to quantize
- * @param step - The step size to quantize to
- * @returns The value rounded down to the nearest step
- */
-const stepFn = (value: number, step: number) => Math.floor(value / step) * step;
+const STEP_SIZE = 100;
 
-/** Default pixel threshold before dimensions update */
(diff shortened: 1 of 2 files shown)


xbtmatt:
tbh I went way overkill with the documentation, I'll trim it down, but it's really not much code, I just wanted to be explicit about what it was doing in case I ever wondered what it was doing again.

```typescript
const stepFn = (value: number, step: number) => Math.floor(value / step) * step;
const STEP_SIZE = 100;

export const useRoughWindowSize = () => {
  const { width, height } = useWindowSize();

  return useMemo(
    () => ({
      width: stepFn(width, STEP_SIZE),
      height: stepFn(height, STEP_SIZE),
    }),
    [width, height]
  );
};
```

in general I try to avoid debouncing because it's inherently an imperative control flow- you're telling the component to *do something* after a certain amount of time. It tends not to play well with state-based checks and forces you to write functions with side effects instead of declarative hooks

It makes sense for things that are inherently time based actions, like user input and things like that, but it can be unintuitive with layout stuff

But idk maybe I'm also just missing an obvious solution here- If you did do it with `useDebounce` I think you'd have to have some sort of setter inside the debounced function, right?

src/typescript/frontend/src/components/misc/background-emojis/use-window-size-with-step.ts
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from "react";
+import useWindowSize from "react-use/lib/useWindowSize";
+
+const STEP_SIZE = 100;
```
xbtmatt, line 4:
```suggestion
const STEP_SIZE = 222;
```

PR #753, opened 2025-03-25
docs(ECO-3128): Add rate limiting flag to the example .env files

CRBl69:
# Description

This PR adds the rate limiting flag to the .env files.


src/typescript/example.mainnet.env
```diff
@@ -46,6 +46,7 @@ NEXT_PUBLIC_DISCORD_METADATA_REQUEST_CHANNEL=""
 HASH_SEED="some random string that is not public"
 GEOBLOCKED='{"countries":[],"regions":[]}'
 NEXT_PUBLIC_CDN_URL=""
+RATE_LIMITING_ENABLED="false"
```
alnoki, line 49: - [ ] Per ECO-3122 explain how this requires the extra service from Vercel marketplace and is only useful in production

src/typescript/example.testnet.env
```diff
@@ -46,6 +46,7 @@ NEXT_PUBLIC_DISCORD_METADATA_REQUEST_CHANNEL=""
 HASH_SEED="some random string that is not public"
 GEOBLOCKED='{"countries":[],"regions":[]}'
 NEXT_PUBLIC_CDN_URL=""
+RATE_LIMITING_ENABLED="false"
```
alnoki, line 49: - [ ] Same as above

alnoki (approved):

xbtmatt (commented):
Agree with Alex here- since this is disabled by default and won't work without a redis/upstash URL/API key, they should all be grouped together with an explanation that they only work if they're all valid and defined

Since it's also network independent, I think it could go in its own `general.env` file or something like that, that way we don't have to repeat it everywhere. We could also move some other env vars to a general env file as well. But that might be something for another PR

Regardless, the flag should be coupled with the URL/API key

xbtmatt (approved):

PR #759, opened 2025-03-26
feat(ECO-3123): Implement rewards remaining tracking

CRBl69:
# Description

- Take into account vault balance.
- Take into account distributed rewards since last querying of remaining
  rewards
- Take into account collected tap out fees since last querying of
  remaining rewards
- Take into account a new melee

# Checklist

### DIFF
--- a/src/typescript/frontend/src/app/arena/page.tsx
+++ b/src/typescript/frontend/src/app/arena/page.tsx
@@ -15,9 +15,16 @@ export const metadata: Metadata = {
 export default async function Arena() {
   if (!FEATURE_FLAGS.Arena) redirect(ROUTES.home);
 
-  const { arenaInfo, market0, market1 } = await fetchCachedMeleeData();
+  const { arenaInfo, market0, market1, rewardsRemaining } = await fetchCachedMeleeData();
 
-  if (!arenaInfo || !market0 || !market1) redirect(ROUTES.home);
+  if (!arenaInfo || !market0 || !market1 || rewardsRemaining === null) redirect(ROUTES.home);
 
-  return <ArenaClient arenaInfo={arenaInfo} market0={market0} market1={market1} />;
+  return (
+    <ArenaClient
+      arenaInfo={arenaInfo}
+      market0={market0}
+      market1={market1}
+      vaultBalance={rewardsRemaining}
+    />
+  );
 }
--- a/src/typescript/frontend/src/components/pages/arena/ArenaClient.tsx
+++ b/src/typescript/frontend/src/components/pages/arena/ArenaClient.tsx
@@ -2,7 +2,6 @@
 
 import type { ClassValue } from "clsx";
 import { Countdown } from "components/Countdown";
-import { FormattedNumber } from "components/FormattedNumber";
 import { useEventStore } from "context/event-store-context/hooks";
 import React, { useEffect } from "react";
 import { createPortal } from "react-dom";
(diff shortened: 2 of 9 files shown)


src/typescript/frontend/src/lib/store/event/event-store.ts
```diff
@@ -173,15 +179,28 @@ export const createEventStore = () => {
...
+                    }
                   } else if (isArenaSwapModel(event)) {
                     melee.swaps.unshift(event);
                   } else if (isArenaCandlestickModel(event)) {
                     handleLatestBarForArenaCandlestick(melee, event);
                   }
+                } else if (isArenaVaultBalanceUpdateModel(event)) {
+                  if (state.vaultBalance !== undefined) {
+                    state.vaultBalance = event.arenaVaultBalanceUpdate.newBalance;
+                  }
```
xbtmatt, line 203: I think we'd rather get the final amount in the writeset changes if possible, otherwise it might be incorrect if the user gets an incomplete set of txns/events after the first one that sets the initial vault balance

src/typescript/frontend/src/components/pages/arena/ArenaClient.tsx
```diff
@@ -48,8 +48,25 @@ const RewardsRemainingBox = ({ rewardsRemaining }: { rewardsRemaining: bigint })
...
+          : s.arenaInfoFromServer.rewardsRemaining;
+      } else {
+        return s.arenaInfoFromServer.rewardsRemaining;
+      }
+    } else {
+      return 0n;
+    }
+  });
+};
+
```
xbtmatt, line 66: Should refactor this to return early rather than use nested if/else

src/typescript/frontend/src/app/arena/page.tsx
```diff
@@ -15,9 +15,16 @@ export const metadata: Metadata = {
 export default async function Arena() {
   if (!FEATURE_FLAGS.Arena) redirect(ROUTES.home);
 
-  const { arenaInfo, market0, market1 } = await fetchCachedMeleeData();
+  const { arenaInfo, market0, market1, vaultBalance } = await fetchCachedMeleeData();
 
-  if (!arenaInfo || !market0 || !market1) redirect(ROUTES.home);
+  if (!arenaInfo || !market0 || !market1 || !vaultBalance) redirect(ROUTES.home);
```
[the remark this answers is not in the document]
xbtmatt: Nice catch bucko

src/typescript/frontend/src/app/arena/page.tsx
```diff
@@ -15,9 +15,16 @@ export const metadata: Metadata = {
...
```
xbtmatt, line 20:
```suggestion
  if (!arenaInfo || !market0 || !market1 || vaultBalance === null) redirect(ROUTES.home);
```

xbtmatt (commented):
I know you're out this week so don't worry about this review- mostly writing it for myself when I refactor it tomorrow

src/typescript/frontend/src/lib/store/event/event-store.ts
CRBl69: This is clientside code, do we ever have access to writesets here ??

CRBl69: But getting it from the txn response will be even more outdated. The vault balance updates very time someone enters. You only get a txn response with the relevant writesets when **you** enter/exit. This means that you will not see it update when needed.

xbtmatt (approved):

PR #761, opened 2025-03-27
chore(ECO-3139): Generate move functions for `favorites` contract

xbtmatt:
# Description

- [x] Auto generate the code using the blueprint tool 

#### Note that this assumes the module address will be the same as the base contract. `market_metadata` also does this, so it should be fine. If you deploy it elsewhere just remind me to update the address in env vars @alnoki


alnoki (approved):

CRBl69 (approved):
beautiful

PR #747, opened 2025-03-24
chore(ECO-3119): Change `paths` aliases in `tsconfig.json` files to conform to TS ecosystem standards

xbtmatt:
# Description

This also lets `simple-import-sort` sort imports more intelligently

The only changes are in the `frontend`, since `sdk` already was using `@/contract-apis`.

The changes made were a simple find and replace on:

- `@containers`   =>   `@/containers`
- `@hooks`        =>   `@/hooks`
- `@icons`        =>   `@/icons`
- `@sdk-types`    =>   `@/sdk-types`
- `@sdk`          =>   `@/sdk`
- `@static`       =>   `@/static`

And then I ran `pnpm run lint:fix` to auto sort imports

EDIT: Just to double check, I also made sure all usages were of the import form with regex. For example for `@containers` I checked that `@containers` appeared the same exact amount of times as `"@containers.*";`, for all replacements above.

PR #764, opened 2025-03-27
emojicoin Version 2  contract 

yue1823:
# Description

Change the Emojicoin symbol limit to 10-32. and rebuild the client of the version 2 contract to make the client support the version 2 Emojicoin. Some functions have emoji bytes that will use the ternary operator to decide which contract to use, but some functions only have a market address, which makes it hard to know if this market is from version 1 or version 2 . Therefore, I use the error callback to fetch form version 2 contract when it fails to fetch version 1.

more detail of the change : `https://docs.google.com/document/d/1yr-BmPs8EcF0jwTbCCA5ItNW2qSdnZyBP6JEu0zt96E/edit?usp=sharing`

# Testing

It works on testnet , and it is successful to create an emoji coin with 10 to 32 bytes. and change the init module to the initialize function, and that will require an admin to run it the first time after publishing. 

testnet address :  `0xed4b36e839063eb381e6dd80b255da59f6abeecb0a707c476c01f9fe1cb15037`

# Checklist

- [ ] Did you check all checkboxes from the linked Linear task? (Ignore if you


alnoki:
@yue1823 first of all thank you for your interest in the emojicoin.fun project!

A 19,000 line PR that affects core logic is a lot to merge in, so I am a bit unsure about the best way to move forward here. For example there are some issues:

1. There are over a dozen merge conflicts (@xbtmatt mentioned this should help: https://github.com/econia-labs/emojicoin-dot-fun/pull/747)
2. There is little division in scope - ideally we'd have reviewable chunks that are at most several hundred lines of code each
3. Full incorporation would require updates to the indexer as well, and e2e tests: https://github.com/econia-labs/aptos-indexer-processors/tree/emojicoin-dot-fun

This is also a substantial feature addition, akin to the emojicoin arena feature, and comes with the some extra maintainer load

I understand that @gregnazario has been advising a bit on this, I think perhaps @xbtmatt can follow up with both of you and figure out a solution here

yue1823:
@alnoki Sorry for this , but most of these codes are a whole copy of the document and the move contract code. @gregnazario told me to create a new folder to store all changed places, not directly change them on it. The real changing place is just around 400-500.

PR #767, opened 2025-03-28
ci(ECO-3153): Update docker test harness to never build docker image in CI tests, always pull

xbtmatt:
# Description

In TypeScript test suites sometimes the runner will try to build a docker image whenever the tags don't line up.

- [x] Update the docker command in the `docker-test-harness.ts` script to never build and always pull.

### DIFF
--- a/src/typescript/sdk/tests/utils/docker/docker-test-harness.ts
+++ b/src/typescript/sdk/tests/utils/docker/docker-test-harness.ts
@@ -104,9 +104,17 @@ export class DockerTestHarness {
 
     const command = "docker";
 
-    const args = ["compose", "-f", LOCAL_COMPOSE_PATH, "--env-file", LOCAL_ENV_PATH, "up"].filter(
-      (arg) => arg !== ""
-    );
+    const args = [
+      "compose",
+      "-f",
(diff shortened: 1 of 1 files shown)


xbtmatt (commented):
Debating if this makes sense or not @alnoki, do you know if this will stop the images with `image: ...` tags in `compose.yaml` from building but not `localnet`, since it's built on the fly?

My main intention here is getting rid of the accidental attempts to build processor/broker in e2e tests, since it can make the tests go on for 10 mins instead of failing early.

alnoki:
> Debating if this makes sense or not @alnoki, do you know if this will stop the images with `image: ...` tags in `compose.yaml` from building but not `localnet`, since it's built on the fly?
> 
> My main intention here is getting rid of the accidental attempts to build processor/broker in e2e tests, since it can make the tests go on for 10 mins instead of failing early.

Hmm I think the `--pull always` could be an issue 

If you want a fast fail for broker/processor, then maybe you could do a command ahead of time that verifiably pulls them? Then when you run the compose command I believe it'll use the local processor/broker images and then will fallback on building localnet

xbtmatt:
Closing as this isn't a huge issue and can be dealt with later if it arises again

PR #768, opened 2025-03-29
fix(ECO-3154): Add `staleTime: TEN_SECONDS` to `useSwapEventsQuery` and `useChatEventsQuery`; refactor `useIsMarketRegistered` `staleTime`

xbtmatt:
# Description

They should be fetched as little as possible. They're fetched every time the window regains focus. The broker/websockets should handle most new incoming events.

I've added `staleTime: TEN_SECONDS` to both of them so that they're not fetched more often than once every 10 seconds. This will still allow users that can't connect over websockets to see new swap/chat events but not every single time the page/component regains focus.

- [x] Also refactored the `staleTime` for `useIsMarketRegistered`, making sure it never refetches unless the number of markets changes. I tested this on a localnet several times to make sure it works.

# Testing

Tested it myself manually by watching the network tab. Ensured each type of request isn't made more than once every 10 seconds.

### DIFF
--- a/src/typescript/frontend/src/components/pages/emojicoin/components/chat/useChatEventsQuery.tsx
+++ b/src/typescript/frontend/src/components/pages/emojicoin/components/chat/useChatEventsQuery.tsx
@@ -10,6 +10,8 @@ import { LIMIT } from "@/sdk/indexer-v2/const";
 
 type ChatEvent = Awaited<ReturnType<typeof fetchChatEvents>>[number];
 
+const TEN_SECONDS = 10000;
+
 export const useChatEventsQuery = (args: z.input<typeof GetChatsSchema>) => {
   const query = useInfiniteQuery({
     queryKey: ["fetchChatEvents", args],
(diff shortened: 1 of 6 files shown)


0xForkh (approved):
Nice!

Would it make sense to have it globally? 
I was also thinking about disabling refetchOnWindowFocus globally.
It can be configured on the queryClient like this:

`const queryClient = new QueryClient({defaultOptions: {queries: {refetchOnWindowFocus: false, staleTime: 10000}}});`

This sets the default behavior for all queries, but queries can later override these default values.

xbtmatt:
@0xForkh  Oh awesome, I was just wondering if there's something for that- I keep writing `staleTime: TEN_SECONDS` everywhere... 😅 

I think `refetchOnWindowFocus: true` is fine as long as it still respects `staleTime: 10000`. That is how it works, right?

Or is there another reason we'd disable refetching on window focus?

0xForkh:
I was not sure how it behaves when refetchOnWindowFocus is true, but you're right, it will only refetch if data is stale. So you can ignore it.

xbtmatt:
Tbh @0xForkh now that I think of it, I think the cognitive overhead of remembering that there's a default 10 second stale time for each query isn't worth saving a couple lines for each file. even redefining `const TEN_SECONDS = 10000` is fine. It's really clear what's happening and no real reason to change it

PR #771, opened 2025-03-29
feat(ECO-3141): Build frontend for Favorite feature

0xForkh:
# Description

- Adds a button component to handle logic to add/remove a market from favorites.
- Adds the button to each market card on home page grid
- Adds the button to the market page

# Todo

- Add a filter on the home page to only show your favorite markets

# Testing

- Navigate to home page or market page, and click on the heart shape button, and approve transaction.

# Checklist


src/typescript/frontend/src/components/favorite-button/favorite-button.tsx
```diff
@@ -0,0 +1,88 @@
...
+      return marketAddress;
+    } else if (emojis) {
+      return getMarketAddress(emojis.map((e) => e)).toString();
+    }
+    console.error("No market address or emojis provided");
+    return "0x0";
+  }, [marketAddress, emojis]);
+
+  const addFavorite = useFavoriteTransactionBuilder(mktAddress, true);
+  const removeFavorite = useFavoriteTransactionBuilder(mktAddress, false);
```
xbtmatt, line 53: Any reason to separate these? Does it not use the right txn builder if the user switches and clicks and it tries to use the current txn builder result?

xbtmatt: I also realize it's a draft so hopefully I'm not breathing down your neck here haha 😅

xbtmatt (commented):
Mentioned in slack but will re-iterate here:

Before this PR is submitted, we need to set the adddress for other networks and/or handle the possibility that it's not set.

In general we should feature flag the whole thing so that every code path with potential "favorites" can swallow any errors (like a missing favorites address or something)

PR #774, opened 2025-04-01
refactor(ECO-3160): Update favorites package to return emoji bytes and have a max limit

xbtmatt:
# Description

The frontend implementation for querying a user's favorites gets fairly involved when there are lots of favorites, since it has to reconcile on-chain data with the database data- particularly when pagination is involved.

- [x] Limit the max number of favorited markets to 100 to simplify the frontend logic necessary
- [x] Add tests for up to and exceeding the max number of favorited markets

It's also much more difficult to properly query markets if we do it by market address.
Thus I've also refactored it to intake a market's emoji bytes (separated per emoji) and output the flattened emoji bytes.

- [x] Replace `market: address` in all functions to use the market's flattened `symbol_bytes: vector<u8>`
- [x] Update tests, ensure they still succeed
- [x] Update `README.md` to remove any inconsistencies

Update TypeScript code
- [x] I also updated the generated TypeScript code for the package.

# Testing

Tests were added for both new refactored characteristics— see above.

```shell
 aptos move test --dev
```

### Register the market
```shell
aptos move run --function-id local::emojicoin_dot_fun::register_market --profile local --args hex:0x04f09f8e83 address:0x1

Transaction submitted: https://explorer.aptoslabs.com/txn/0xf79afbe06a031d7097e9fa084dea32d1de0be936fdf8c2b615c282186fae7618?network=local
```

```json
{
  "Result": {
    "transaction_hash": "0xf79afbe06a031d7097e9fa084dea32d1de0be936fdf8c2b615c282186fae7618",
    "gas_used": 513157,
    "gas_unit_price": 100,
    "sender": "111f39f15da7e418cbb2188666eedd8e9bade9dae11cf949b0736c848025c111",
    "sequence_number": 2,
    "success": true,
    "timestamp_us": 1743475636571111,
    "version": 953,
    "vm_status": "Executed successfully"
  }
}
```

### Adding it as a favorite
```shell
aptos move run --function-id local::emojicoin_dot_fun_favorites::add_favorite --profile local --args hex:0x04f09f8e83

Transaction submitted: https://explorer.aptoslabs.com/txn/0x79ca8f7068e53d9231a302fa7ccc76aa082654a56cfac5c0f12d817acb9c2779?network=local
```

```json
{
  "Result": {
    "transaction_hash": "0x79ca8f7068e53d9231a302fa7ccc76aa082654a56cfac5c0f12d817acb9c2779",
    "gas_used": 457,
    "gas_unit_price": 100,
    "sender": "111f39f15da7e418cbb2188666eedd8e9bade9dae11cf949b0736c848025c111",
    "sequence_number": 3,
    "success": true,
    "timestamp_us": 1743475643642858,
    "version": 970,
    "vm_status": "Executed successfully"
  }
}

```

### View my new favorites
```shell
aptos move view --function-id local::emojicoin_dot_fun_favorites::view_favorites --args address:local --profile local
```

```json
{
  "Result": [
    [
      "0xf09f8e83"
    ]
  ]
}
```

### DIFF
--- a/src/move/favorites/sources/emojicoin_dot_fun_favorites.move
+++ b/src/move/favorites/sources/emojicoin_dot_fun_favorites.move
@@ -233,7 +233,7 @@ module favorites::emojicoin_dot_fun_favorites {
             let emoji_bytes = *vector::borrow(&emojis, i);
             add_favorite(&account_1_signer, emoji_bytes);
         };
-        
+
         let num_favorites = view_favorites(ACCOUNT_1).length();
         assert!(num_favorites == MAX_NUM_FAVORITES);
     }


src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -19,17 +19,21 @@ module favorites::emojicoin_dot_fun_favorites {
     const E_ALREADY_FAVORITE: u64 = 2;
     /// Market is not a favorite.
     const E_NOT_FAVORITE: u64 = 3;
+    /// User already has the maximum number of favorites.
+    const E_USER_HAS_MAX_NUMBER_OF_FAVORITES: u64 = 4;
+
+    const MAX_NUM_FAVORITES: u64 = 100;
```
alnoki, line 25: Since this is an `OrderedMap` all queries are going to hit a single resource, which I believe starts to get linearly more expensive with each extra byte once you pass the 4K threshold

Luckily the flattened emoji byes should be at max 10 bytes, compared with 32 in an address, so at most 1k bytes for all the lookup keys. Since `Nil` has no bytes this should basically still be fine with the "free" 4k bytes in a storage slot

I'd even be fine with lowering it to 50 or 25, because 100 is choice overload from a cognitive load standpoint

But we can treat people like adults and let them have 100 favorites if they really want

Feel free to resolve this comment yourself

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
```diff
@@ -9,7 +9,7 @@ module favorites::emojicoin_dot_fun_favorites {
     #[event]
     struct Favorite has copy, drop, store {
         user: address,
-        market: address,
+        symbol_bytes: vector<u8>,
```
alnoki, line 12: (Leaving this comment here to block merge without leaving rejection review)

- [ ] Update README

alnoki (approved):

alnoki:
Also just flagging I changed title from ~`contract`~ to `favorites`

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
xbtmatt: Lowered it to 25! You have a good point- I was going to ask you if we could lower it so I totally agree 😁

src/move/favorites/sources/emojicoin_dot_fun_favorites.move
xbtmatt: Done!

PR #435, opened 2024-12-02
[ECO-2537] Add parcel cache implementation

CRBl69:
# Description

This PR adds the parcel cache implementation.

PR #779, opened 2025-04-04
feat(ECO-2537-rev2): Implement new parcel caching

0xForkh:
# Description

This PR copies over the parcel implementation from @CRBl69's #435 to main branch.
Trade history and chat have changed a lot since the original PR was opened, so I excluded these changes.
This PR focuses on adding the parcel cache implementation and applying it to the candlestick route.

Can be added to Chat, Trade history and Arena candlesticks in a separate PR.

# Checklist


src/typescript/frontend/src/utils/index.ts
```diff
@@ -3,7 +3,6 @@ import { sleep } from "@/sdk/utils";
 
 export { isDisallowedEventKey } from "./check-is-disallowed-event-key";
 export { checkIsEllipsis } from "./check-is-ellipsis";
-export { getEmptyListTr } from "./get-empty-list-tr";
```
0xForkh, line 6: This export was preventing some tests from running. Same for the broker-v2 export

CRBl69 (changes requested):
Thank you for updating it.

There seem to be a problem with the keys.

For example, when I run the code, the historic threshold is `1744018216560` but the end key is `1744200000`. This is probably due to incorrect params in the parcel builder (line 51 and 54 in the candlestick route).

also, we might want to reduce the normal cache duration for 15s candlesticks.

CRBl69:
That seems to be an error with the original code. The best fix would be to use seconds everywhere rather than milliseconds.

PR #780, opened 2025-04-04
chore(ECO-3049): Add search param validation to all routes in /api/**/route.ts

0xForkh:
# Description
- Adds zod schema for symbol emojis
- Adds zod schema for AccountAddress
- Adds params validation with zod schema for pool api
- Improves client side fetching of pool data with useQuery
- Adds `api-route-error-handler.ts` which can be used to wrap api endpoints. This allows to catch errors without having to add try-catch clauses to every individual api endpoints. If the error is a validation error, it will send a proper error response that describes each invalid param.

# Todo
- Add validation for remaining routes
- Wrap remaining routes with `api-route-error-handler.ts`

# Checklist
No checklist

### DIFF
--- a/src/typescript/sdk/src/utils/validation/account-address.ts
+++ b/src/typescript/sdk/src/utils/validation/account-address.ts
@@ -1,0 +1,9 @@
+import { AccountAddress } from "@aptos-labs/ts-sdk";
+import { z } from "zod";
+
+export const AccountAddressSchema = z
+  .string()
+  .refine((arg) => AccountAddress.isValid({ input: arg }).valid, {
+    message: "Invalid account address format",
+  })
+  .transform((val) => AccountAddress.from(val));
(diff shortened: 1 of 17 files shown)


src/typescript/sdk/src/utils/validation/account-address.ts
```diff
@@ -0,0 +1,9 @@
+import { AccountAddress } from "@aptos-labs/ts-sdk";
+import { z } from "zod";
+
+export const AccountAddressSchema = z
+  .string()
+  .refine((arg) => AccountAddress.isValid({ input: arg }).valid, {
+    message: "Invalid account address format",
```
xbtmatt, line 7: make sure to check that the `arg` is actually defined, because `isValid` doesn't check

0xForkh:
@xbtmatt Any idea why this is failing in CI but it's not picked up locally?
image

src/typescript/sdk/src/utils/validation/account-address.ts
0xForkh: I added some checks in f7c0a074c0b6b6422a73a701e8711587c0abb31a

It seems to work properly. When I pass a null or undefined value, and the schema isn't explicitely marked as optional, it will throw an error.
And when it's marked as optional, it will not throw an error and return undefined instead.

Seems to work as intended

PR #781, opened 2025-04-05
refactor(ECO-3157): Revamp user profile and history tab

xbtmatt:
# GitHub issue

This PR appeared to merge per https://github.com/econia-labs/emojicoin-dot-fun/commit/52d8d49cdacb943333e9373615134a7f0d08a951, even though it did not close this PR

# Revamp user profile and history tab

- [x] Ensure primary breakdown is accurate
  - [x] Fetch with query, keep up to date with indexer.
  - [x] Wait on submitted txn version.
  - [x] Keep it simple - no client side calculations/state management unless necessary, just fetch from indexer
- [x] Ensure individual breakdowns are accurate using the same method as above
  ~- [ ] There are still issues with the indexer reporting incorrect values though. For example, one of the melees I triggered somehow resulted in a profit of `1.12e50` APT or something absurd~ tracked in another issue
- [x] Ensure all melees show up in "History" properly
  - [x] This is generally as simple as tracking a user's open escrows/positions
  - [x] Queryable from indexer as `/positions/[user]` and `/historical-positions/[user]`
- [x] Ensure `Position` tab is up to date with user interactions client side
  - [x] Update upon submitting transaction- check from indexer as well.
  - [x] Use `useCurrentPositionQuery` as well as open escrows in state?
- [x] Fix the `row.exited = true` . ctrl + f "Fix this ASAP"
- [x] Fully remove the rest of the `setPosition` state setter defined in `ArenaClient` at the top component
- [x] Use the `useArenaEscrowStore` hook to get the partial value of the latest escrow value in state. Can use the position query for the rest of the pnl reports/stats
  - [x] Keep it simple by *only* using the position for stats, and using the escrow hook everywhere else.
- [x] Ensure no infinite rerenders/cyclical deps somehow
- [x] Ensure fetches are deduped properly (5-6 flurry of them on historical positions query)
- [x] add &minimumVersion= to the historical positions and current position queries. it's set up in route.ts but not being used yet.
- [x] Leave minimum version in, but don't use it. All position stats can be calculated entirely client side just mimicking Bogdan's position diff model.
- [x] Make sure to remove the refetch in `useCurrentPositionQuery` and generally everywhere
- [x] Replace `useCurrentPositionQuery` with the coalesced version of the query (with the escrow in state) and rename the existing query hook to `useCurrentPositionQueryInternal`
- [x] Ensure the user must lock in if they're currently locked in (otherwise they'll get a txn error in simulation)

# Testing

Tested manually an inordinate amount of times.


src/typescript/frontend/src/components/pages/arena/tabs/enter-tab/EnterTabLockPhase.tsx
```diff
@@ -39,21 +45,22 @@ export const EnterTabLockPhase: React.FC<{
...
         </div>
         <div className="flex uppercase justify-between text-2xl text-light-gray py-[0.8em] mx-[0.8em] border-dashed border-b-[1px] border-light-gray ">
           <div>Match amount</div>
-          <FormattedNumber
-            value={innerLock ? BigInt(Math.floor(Math.min(5 * 10 ** 8, Number(amount / 2n)))) : 0n}
-            nominalize
+          <FormattedNominalNumber
+            value={
+              lockedInToggle ? BigInt(Math.floor(Math.min(5 * 10 ** 8, Number(amount / 2n)))) : 0n
+            }
```
xbtmatt, line 63: I think this is incorrect- should be updated in another PR? It's dynamic based on the time of the melee

src/typescript/frontend/src/lib/queries/latest-emojicoin-version.ts
```diff
@@ -0,0 +1,51 @@
...
+      });
+
+    if (latestVersion >= minimumVersion) return true;
+
+    await sleep(FETCH_INTERVAL);
+  }
+
+  console.trace(`Failed to wait for version ${minimumVersion} after ${maxTries}`);
+  return false;
+};
```
xbtmatt, line 51: Note that this isn't necessary anymore, but it's nice to have the code in case we need it at some point. I got it to work properly with caching and proper serialization/deserialization, but didn't need it anymore because I added a mini processor to the client, so we don't have to refetch position stats anymore after the initial load.

src/typescript/frontend/src/lib/store/arena/escrow/hooks.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  useEffect(
+    () =>
+      globalUserTransactionStore.subscribe(
+        (state) => (transactionsRef.current = address && state.addresses.get(address)),
+        (newTransactions, _prevTransactions) => {
+          if (address && newTransactions) {
+            globalEscrowStore.getState().processTransactions(...newTransactions);
+          }
+        }
```
xbtmatt, line 67: ideally we'd dedupe these- right now it re-processes all transactions.

It's not a big deal because no user will ever submit more than 100-200 txns in a single session, and if they did it wouldn't really be a huge deal anyway, because that's still a relatively small amount of txns to process.

alnoki (approved):

src/typescript/frontend/src/components/pages/arena/ArenaClient.tsx
```diff
@@ -113,58 +106,59 @@ const Mobile = (props: ArenaPropsWithPositionHistoryAndEmojiData) => {
...
+// Necessary to add a display name because of the React.memo wrapper.
+Mobile.displayName = "Mobile";
+
+/**
+ * When a melee ends, all clients will refresh at once. To avoid hammering the indexer in case of huge traffic, stagger
+ * the requests by randomly delaying them for 1-5 seconds. Getting the new melee data is a once a day thing, so it's
+ * fine if this is delayed by a few seconds for a client.
+ *
+ * In a development environment, this is just a flat, shorter amount of time.
+ */
```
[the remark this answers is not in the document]
xbtmatt: I said `indexer` but sort of just meant the website in general- I figured if there's ever a huge amount of traffic it's probably better off slightly staggered rather than concentrated all at once. I'm not sure if it *is* cached properly, and don't know if it will be by the time we launch- so this was simple/easy to add in for the time being. Pretty easy to remove it and it's a few seconds delay once a day, so I figured it wasn't much of an issue

It would be great if you could double check the indexes for these fetches, though

WRT not querying a user's position conditionally- true! But adding that logic to the useQuery hook for the current position is a bit more complicated than that- if you want to figure out a way to ensure it doesn't call it on reload that'd be great too

CRBl69 (approved):
L.G.T.M. stands for Let's Go Turbo Merge

src/typescript/frontend/src/components/pages/arena/tabs/enter-tab/EnterTabLockPhase.tsx
xbtmatt: Filed an issue for it

alnoki:
This PR appears to have already merged: https://github.com/econia-labs/emojicoin-dot-fun/commit/52d8d49cdacb943333e9373615134a7f0d08a951

PR #783, opened 2025-04-05
fix(ECO-3170): Fix swaps incorrect sorting; update chat sorting

xbtmatt:
# Description

- [x] It was sorting by a string rather than by a number-like value in the swaps history.
- [x] Make them both consistent and correct and use the market nonce.


xbtmatt:
@0xForkh have the swaps animations been looking like this? They look different. It doesn't look like it's animating properly

alnoki (approved):

PR #785, opened 2025-04-07
test(ECO-3175): Add E2E test with processor restarts

CRBl69:
# Description

This PR adds E2E tests that include processor restarts to test the initialization of data in the processor.

# Checklist

### DIFF
--- a/src/typescript/sdk/tests/e2e/arena/restart.test.ts
+++ b/src/typescript/sdk/tests/e2e/arena/restart.test.ts
@@ -1,23 +1,20 @@
+import type { MeleeEmojiData } from "../../../src";
 import {
   fetchArenaMeleeView,
   fetchMeleeEmojiData,
-  MeleeEmojiData,
   sleep,
   toArenaLeaderboardHistoryModel,
 } from "../../../src";
 import { EmojicoinClient } from "../../../src/client/emojicoin-client";
-import {
-  postgrest,
-  TableName,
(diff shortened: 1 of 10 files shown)


alnoki (changes requested):
- [x] Update version binaries across compose and stack deploy files for new processor
- [x] Update submodule with release binary once built

src/typescript/sdk/tests/e2e/arena/restart.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    await sleep(10000);
+
+    // Crank
+    const res = await emojicoin.arena.enter(account, 1n, false, melee.market0.symbolEmojis, melee.market1.symbolEmojis, "symbol0");
+
+    await waitForProcessor(res);
+
+    const leaderboard = await postgrest
+      .from(TableName.ArenaLeaderboardHistory)
+      .select("*")
```
xbtmatt, line 54: currently this fails (as expected) With the wrong version, test that it fails before testing that it succeeds in the new image

xbtmatt: tested, and it works as expected!

xbtmatt (approved):
Awesome, thanks for adding this test to confirm. it's nice to have a way to restart the processor in tests now

alnoki (approved):

PR #787, opened 2025-04-09
feat(ECO-3186): UI improvements on arena page

0xForkh:
# Description

edit by matt: see #819 for the description of extra fixes on top of this

- [x] new buttons variants: `solid` and `border`
- [x] anchor arena/position buttons to the bottom
- [x] swap color scheme for mobile menu
- [x] arena animations during interactions -> Will show loading animation when status is in "pending" state
  - Should this also appear during the "prompt" status?
# Checklist

image

image


src/typescript/frontend/src/components/header/components/mobile-menu-item/mobile-menu-item.tsx
```diff
@@ -0,0 +1,52 @@
+import Arrow from "components/svg/icons/Arrow";
```
0xForkh, line 1: Simplified this into a single file. Makes it easier to understand instead of having to visit 3 different files.

src/typescript/frontend/src/components/text/index.tsx
```diff
@@ -25,7 +25,7 @@ const wordBreak = ({ wordBreak }: ThemedProps) => {
 const Text = styled.p.attrs<TextProps>(({ textScale = "display6" }) => ({
   textScale,
 }))`
-  color: ${({ theme, color }) => (color ? theme.colors[color] : theme.colors.white)};
+  color: ${({ theme, color }) => (color ? theme.colors[color] : "inherit")};
```
0xForkh, line 28: White is already the default color based on global css. No need to override here. Setting this to inherit allows to override the color in one of the parents of this component without having to pass it explicitely as a prop of this item.

src/typescript/frontend/src/components/svg/icons/Planet.tsx
```diff
@@ -1,19 +1,15 @@
...
-          fill={darkTheme.colors[color]}
```
0xForkh, line 15: I just removed the explicit fill here because it overrides the fill supplied by css which makes it confusing. I checked everywhere to make sure it didn't break anything.

xbtmatt (commented):
Awesome, thanks for the cleanup and refactors btw. Only thing is the animation for the rocket is pretty short, if you want to remove it from this PR and add it later we can get this in right away- I think it might take a few iterations to get right?

Unless you think you can fix it quickly. But basically it's just showing up for like...0.2 seconds

xbtmatt (commented):
Also the buttons on the position page are a bit low- any way we can get a solid in between? Haha. more like 20-30% up rather than at the very bottom. I think the figma shows an example of it, lmk if you can't find it

0xForkh:
@xbtmatt I pushed the changes to #805 since a lot of stuff from this PR is overriden there.
Increased padding-bottom from 4 to 7: 14b27d64d2202a111d20a99dac9a945df7af0939
Set minimum duration for loading to 2.3 seconds: 345b19b99cc3b32c1322ffa580a63f284b3b2b86

xbtmatt (commented):
There were some custom styles applied in the mobile menu for connect wallet to ensure it lines up horizontally with the other items, could you get it lined up again? 😁 

old (mostly lined up):
image

new (not lined up):
image

Also, I think the chat page also still has a grey background (edit: it's only when you're on mobile sizes, aka below 748px or whatever)

Also, the new animation is great lol, I might add some fluidity to it with framer-motion, but it's pretty cool

xbtmatt (approved):
Looks good to merge, I addressed most outstanding nits/issues myself just to speed up getting this in in #819 

Should be ready to go!

alnoki (approved):

PR #795, opened 2025-04-10
refactor(ECO-3177): Use new indexer SDK

CRBl69:
This PR moves the indexer sdk in the mono repo, and adapts it to work with the new indexer SDK.

The only real changes can be found in the `src/processor.rs` file. I recommend running a diff (using the standard UNIX `diff` tool) between this file, and the file in the previous indexer which is located at `rust/processor/src/processors/emojicoin_dot_fun/processor.rs`.

Some other changes include:

- A new main file to start the processor: `src/main.rs`.
- Removal of all Aptos related tables, and inclusion of the `processor_status` table in migrations, to keep backwards compatibility.
- Updating of the docker file (I recommend running a diff as for the `processor.rs` file).
- Updating of the `start.sh` file to update to a compatible config.
- Addition of the processor crate to the parent `Cargo.toml`.
- Addition of `push-processor.yaml` GitHub Action.


alnoki (approved):
Approving for line number override, will defer to @xbtmatt on review and also to loop me in as needed

src/docker/compose.yaml
```diff
@@ -83,7 +83,7 @@ services:
     depends_on:
       postgres:
         condition: 'service_healthy'
-    image: 'econialabs/emojicoin-dot-fun-indexer-processor:7.0.4'
+    image: 'econialabs/emojicoin-dot-fun-indexer-processor:10.0.0-rc.1'
```
alnoki, line 86: Why `10` instead of `8`?

CRBl69: I wanted to leave room, in case we don't merge this PR instantly and we push other versions first. I thought getting this to work would take more time, and that's why I did that in the first place.

alnoki: Understood - I'll leave this comment unresolved until this is ready for a merge, then, and we can rebuild a proper version right at the end

PR #800, opened 2025-04-11
chore(ECO-3213): Update wallet adapter packages and manage breaking changes accordingly

xbtmatt:
# Description

The wallet adapter has a new major version and it fixes the *really* cumbersome issue of disconnecting on a local network! This issue has cost me dozens of hours probably so I'm very happy to see that the newest version fixes it.

I've bumped the versions fixed all breaking changes accordingly.

Since the new wallet adapter version only accepts wallets that conform to the AIP 62 wallet adapter standard, only 3 wallets (not counting the Google/Apple OAuth AptosConnect wallets) remain (Petra, Nightly, and OKX). That's okay though, because the others became difficult to use with outdated dependencies and inconsistencies in how to submit txns.


xbtmatt:
Note that the wallet adapter versions were bumped very recently and I think they forgot to bump @aptos-labs/derived-wallet-base, but they did bump the solana/ethereum derived wallet packages, so there's a mismatch and that's why the build is failing.

I imagine it'll be fixed relatively soon- when it is I'll bump the packages and update accordingly!

PR #805, opened 2025-04-13
feat(ECO-2323): Add tabs UX/UI to the emojicoin page

0xForkh:
# Description
- Add new design for tabs: https://www.figma.com/design/gPCK4ByZV718N8GPd7MPm0/Econia-Labs-%7C-emojicoin.fun?node-id=15869-1526&p=f&t=OPdjA0H3FiLidbQN-0
- Update market page with new tabs
- Update portfolio page with new tabs
- Update arena page with new tabs
- Add a workaround to move the scollbar to the top of the tabs
- Some cleanup in the arena tabs content to make it consistent

# Testing
- Visit pages /arena, 

# Notes
- I took some liberties for some of the designs. For example the background of the tabs content on the Arena page is now a different tint to match with the tint of the selected header. Some of these designs are to be discussed

# Checklist
No checklist

# Screenshots
image
image
image
image
image

### DIFF
--- a/src/typescript/frontend/package.json
+++ b/src/typescript/frontend/package.json
@@ -107,7 +107,7 @@
     "check:ci": "tsc -p tests/tsconfig-ci.json",
     "check:tests": "tsc -p tests/tsconfig.json",
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist && rm -rf .next",
-    "dev": "NODE_OPTIONS='--inspect' next dev --turbo -H 0.0.0.0 --port 3001",
+    "dev": "NODE_OPTIONS='--inspect' next dev --turbo --port 3001",
     "format": "pnpm _format --list-different --write",
     "format:check": "pnpm _format --check",
     "lint": "eslint --max-warnings=0 -c .eslintrc.js --ext .js,.jsx,.ts,.tsx .",
--- a/src/typescript/frontend/package.json
+++ b/src/typescript/frontend/package.json
@@ -107,7 +107,7 @@
     "check:ci": "tsc -p tests/tsconfig-ci.json",
     "check:tests": "tsc -p tests/tsconfig.json",
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist && rm -rf .next",
-    "dev": "NODE_OPTIONS='--inspect' next dev --turbo --port 3001",
+    "dev": "NODE_OPTIONS='--inspect' next dev --turbo -H 0.0.0.0 --port 3001",
     "format": "pnpm _format --list-different --write",
     "format:check": "pnpm _format --check",
     "lint": "eslint --max-warnings=0 -c .eslintrc.js --ext .js,.jsx,.ts,.tsx .",
--- a/src/typescript/frontend/src/components/inputs/search-bar.tsx
+++ b/src/typescript/frontend/src/components/inputs/search-bar.tsx
@@ -36,14 +36,14 @@ const Border = styled(Flex)`
   }
 `;
 
-const SearchBar = () => {
+const SearchBar = ({ className }: { className?: string }) => {
   const setMode = useEmojiPicker((state) => state.setMode);
   useEffect(() => {
     setMode("search");
     /* eslint-disable-next-line react-hooks/exhaustive-deps */
   }, []);
   return (
-    <Container>
+    <Container className={className}>
       <Border>
         <div
           style={{
--- a/src/typescript/frontend/src/components/pages/arena/ArenaClient.tsx
+++ b/src/typescript/frontend/src/components/pages/arena/ArenaClient.tsx
@@ -3,6 +3,7 @@
 import type { ClassValue } from "clsx";
 import { Countdown } from "components/Countdown";
 import { useEventStore } from "context/event-store-context/hooks";
+import { cn } from "lib/utils/class-name";
 import React, { useEffect } from "react";
 import { createPortal } from "react-dom";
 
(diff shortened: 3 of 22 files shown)


xbtmatt (commented):
Looks so much nicer overall, thanks for doing this! Few things, mostly visual things:


#### 1. Background on arena tabs
  - I saw the PR description and yes I do think the new gray background contrasts too much for it to fit visually- is it difficult to get the arena tabs to remain the same dark black but have the tabs be the highlighted color, like the trade history?
  - I thought it looked ok at first but after looking at it more it looks a bit out of place.

#### 2. Emojis on tabs aren't using the emoji class
  - I think you need to wrap the emoji displays with the <Emoji ... /> component we have for styling basic emojis with the properly rendered font. The emojis I see on arena now are my default windows emojis, not the Noto ones

#### 3. The font sizing of the new tabs is much smaller than before

  - Did you change it for a reason or was this an accidental change? Just curious, because I'm wondering if you changed it to make the tabs fit horizontally or it was just an accident

#### 4. The progress/rocket thing is no longer centered on the main view ("Pick your side" tab)
  - It's also a bit larger than before

#### 5. The horizontal scroll bar seems to show up (on the market page) even when the content fits and doesn't need to scroll

- This is max width on my 4k monitor, so it definitely doesn't need any extra room

image

#### 6. Also, what exactly is this class? Do `*:grow` and `*:basis-0` actually do something?

I couldn't find documentation on this syntax online, but only briefly looked bc copilot commented on it

```jsx
<div className="px-2 w-[100%] text-white flex items-center *:grow *:basis-0">
```
  
#### 6. Unrelated to this PR, but...

- could you update the link in the "Info" tab to be properly hyperlinked, rather than just text in parantheses. It just says `(https://discord.gg/econia)`

xbtmatt (changes requested):
Thanks for this! It already looks so much better. There's several visual regressions here though, so I want to touch on that really quick before listing them.

In general, I've noticed a few PRs lately have had these small visual regressions that probably could've been caught with a quick pass before opening them up for review.

A side-by-side check between `main` and your PR — just clicking around and looking for differences — would catch most of these and really cut down on review time and async iteration.

I know it's a bit of a pain (ideally we'd automate it somehow) but it really helps on my end — otherwise I have to spend time discerning which changes were intentional and which were accidental. If I can count on that being done before I jump in, it'd make the process a lot smoother and more efficient.

I really do appreciate your efforts in cleaning these up and all these updates overall, though. You're doing most of them without much guidance and navigating the mess pretty well, so I just want to make it clear that overall it's been a massive help!

Now for the things I noticed:

---

#### 1. Background on arena tabs
  - I saw the PR description and yes I do think the new gray background contrasts too much for it to fit visually- is it difficult to get the arena tabs to remain the same dark black but have the tabs be the highlighted color, like the trade history?
  - I thought it looked ok at first but after looking at it more it looks a bit out of place.

#### 2. Emojis on tabs aren't using the emoji class
  - I think you need to wrap the emoji displays with the <Emoji ... /> component we have for styling basic emojis with the properly rendered font. The emojis I see on arena now are my default windows emojis, not the Noto ones

#### 3. The font sizing of the new tabs is much smaller than before

  - Did you change it for a reason or was this an accidental change? Can't really tell

#### 4. The progress/rocket thing is no longer centered on the main view ("Pick your side" tab)
  - It's also a bit larger than before

#### 5. The horizontal scroll bar seems to show up (on the market page) even when the content fits and doesn't need to scroll

- This is max width on my 4k monitor, so it definitely doesn't need any extra room

image

#### 6. Also, what exactly is this class? Do `*:grow` and `*:basis-0` actually do something?

I couldn't find documentation on this syntax online, but only briefly looked bc copilot commented on it

```jsx
<div className="px-2 w-[100%] text-white flex items-center *:grow *:basis-0">
```

0xForkh:
Thanks for the review 🙏🏻
Yes you're right, I should check more carefully before submitting. 
Also some of these issues are due to browser or OS differences, so I will create a checklist where I test on every browser and both on windows and mac before submitting in the future.
I will also make sure to list all the intentional changes to avoid any confusion in the future.

1. It's not really a technical constraint, but for design consistency. This is what it looks like if the tab is selected but the background of the content is black:
image
But since you don't like the gray background, we can instead change the active tab background and go with something like this:
image

3. This is on purpose, I made it smaller to keep it consistent with tabs on other pages. To keep it consistent, I had 2 options, either I make it big everywhere, or small on arena page. Making it bigger on market page would make it take too much space and it didn't look great, so I went with option 2. If consistency is not an issue, I can make it bigger on arena page and keep a smaller size on other pages. wdyt?

5. I can't reproduce on my mac, I will try on windows.

6. The `*` is a selector used to apply style to every direct child of the element. 
So you can do this:
```html
<div className="*:grow *:basis-0">
  <div>1</div>  
  <div>2</div>
  <div>3</div>
</div>
```
Instead of:
```html
<div>
  <div className="grow basis-0">1</div>
  <div className="grow basis-0">2</div>
  <div className="grow basis-0">3</div>
</div>
```
As for the purpose of grow and basis-0, they are both part of flexbox, and in this case they allow each child item to take an equal amount of space.

Also as a side note, the reason I replaced a lot of `h-[100%]` with `grow` is because setting height to 100% doesn't work when the item has siblings. `h-[100%]` will always try to match the parent's height, even if there are siblings taking up some space which leads to a lot of headaches trying to figure out why the content is overflowing. 
Whereas `grow` will fill the remaining space in the container, and never overflow. 
It's also a great way to push stuff to the bottom of a container. If there are two divs, and you add grow to the first div, it will expand and push the second div to the bottom of the container.

I will ping you after fixing and testing everything again, thanks again for the review 🙏🏻.

0xForkh:
For point 4. even on my windows I wasn't able to reproduce. But I changed the logic a bit and normally it shouldn't render the container with the scrollbar at all if the content is not overflowing. Hopefully it fixes the issue.

PR #807, opened 2025-04-14
refactor(ECO-3225): Use the same asset IDs for the asset route as for the pair route for GeckoTerminal

CRBl69:
# Description

This PR updates the asset route for GeckoTerminal to use the same IDs as the pair route.

# Testing

See vercel at `/api/geckoterminal/asset?id=ADDRESS::coin_factory::Emojicoin`.

# Checklist


src/typescript/frontend/src/lib/dexscreener/routes.ts
```diff
@@ -25,17 +26,33 @@ import {
...
-  const marketEmojiData = toMarketEmojiData(assetId);
-  const symbolEmojis = symbolEmojiStringToArray(assetId);
-  const marketState = await fetchMarketState({ searchEmojis: symbolEmojis });
+
+  let marketState: MarketStateModel | null;
+
+  if (ops.geckoTerminal) {
+    if (!/^0x[a-f0-9]{64}::coin_factory::Emojicoin$/.test(assetId)) {
+      return new NextResponse("id must be a valid asset ID", { status: 400 });
+    }
```
xbtmatt, line 43: Can add a better error response here to save some back and forth if they don't format it properly

Something like: `Asset ID must follow the form: 0x[a-f0-9]{64}::coin_factory::Emojicoin` even

PR #812, opened 2025-04-15
fix(ECO-3224): Update the link in the arena Info tab to be properly hyperlinked

0xForkh:
# Description

Fixes the text and makes the link clickable
image

# Testing

Visit /arena page, select info tab and scroll to bottom


xbtmatt (approved):
I updated this not to use dangerouslySetInnerHTML- wasn't a big deal but figured we don't need the type definition for `Info` anyway, so I just put the JSX directly in the paragraph element

PR #816, opened 2025-04-16
fix(ECO-3246): Fix/update pre-commit GitHub Action caches

xbtmatt:
# Description

`pre-commit` seems to be failing consistently lately in CI due to an invalid venv.

See this failed run: https://github.com/econia-labs/emojicoin-dot-fun/actions/runs/14499155811/job/40676172151?pr=807

I believe the main issue was disparate python versions, which I fixed by pinning the default python version in pre-commit-config.yaml. This also consolidates the pre-commit caches (there were multiple caches with multiple python versions being used, one for each version used in different pre-commit hooks)

I also:

- [x] Updated the cache key for the python dependencies to include more info like os and python version, and only run cache save on cache miss.
- [x] Removed `Cache Poetry Install` job, since it never actually cached anything.
The path used wasn't valid `path: '${{ runner.home }}/.local/src/python/hooks'`.
I believe the poetry install is already cached by the `snok/install-poetry` action.
Proof: old logs output `Cache not found for input keys: poetry-1.8.2`, see here for example https://github.com/econia-labs/emojicoin-dot-fun/actions/runs/14143029337/job/39626956266
- [x] Updated the old node/pnpm installation processes. The pnpm installation process is now in its own action in our repo with proper deps caching, so I updated pre-commit to use that instead.
- [x] Updated the caching actions versions to @v4 and pre-commit action patch version
- [x] Also ran `pre-commit autoupdate --repo https://github.com/<REPO>` for `pre-commit/pre-commit-hooks` and `scop/pre-commit-shfmt` per the warning in `pre-commit/action` logs

#### Fixed the below warnings by running the pre-commit autoupdate command suggested in them:
[WARNING] Hint: often `pre-commit autoupdate --repo https://github.com/scop/pre-commit-shfmt` will fix this.
[WARNING] Hint: often `pre-commit autoupdate --repo https://github.com/pre-commit/pre-commit-hooks` will fix this.

#### Fixed by only conditionally running cache save on cache miss:
```shell
Run actions/cache/save@v3
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/emojicoin-dot-fun/emojicoin-dot-fun --files-from manifest.txt --use-compress-program zstdmt
Failed to save: Unable to reserve cache with key pydeps-9a660eeadb5503d2437d6cbfd092308a353f7a4bc8e25352df0ce76f892d6e4f, another job may be creating this cache.
Warning: Cache save failed.
```

# Testing

Should succeed in CI.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -14,38 +14,29 @@ jobs:
     steps:
     - uses: 'actions/checkout@v4'
     - run: 'git submodule update --init --recursive src/rust/processor'
-    - uses: 'actions/setup-python@v3'
+    - id: 'setup-python'
+      uses: 'actions/setup-python@v5'
       with:
         python-version: '${{ env.PYTHON_VERSION }}'
     - name: 'Install libdw-dev'
       run: |
         sudo apt-get update
         sudo apt-get install -y libdw-dev
-    - uses: 'actions/setup-node@v4'
-      with:
-        node-version-file: '${{ env.TS_DIR }}/.node-version'
-        registry-url: 'https://registry.npmjs.org'
-    - uses: 'pnpm/action-setup@v4'
-      with:
-        package_json_file: '${{ env.TS_DIR }}/package.json'
-    - run: 'cd ${{ env.TS_DIR }} && pnpm i'
-    - name: 'Cache Poetry Install'
-      uses: 'actions/cache@v3'
-      with:
-        key: 'poetry-${{ env.POETRY_VERSION }}'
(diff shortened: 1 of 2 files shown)


alnoki (approved):
Thanks for doing this

I had in my own notes to add a generic `pre-commit` meta-runner to the `common` repo so that can handle all this overhead caching, that way we can just add individual `pre-commit` hooks as their own actions rather than as one giant block of different hooks

I just filed a reference to this on ECO-3006 accordingly

PR #822, opened 2025-04-19
fix(ECO-3251): Fix arena section on home page on mobile device

0xForkh:
# Description

There was a max height css property that would create overlapping issues on mobile devices.
I removed it completely since it didn't seem to have any effect other than breaking the layout on mobile. 

# Testing

The bug can be reproduced on desktop by increasing the browser scaling to a high value (175-200%)

### DIFF
--- a/src/typescript/frontend/package.json
+++ b/src/typescript/frontend/package.json
@@ -108,6 +108,7 @@
     "check:tests": "tsc -p tests/tsconfig.json",
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist && rm -rf .next",
     "dev": "NODE_OPTIONS='--inspect' next dev --turbo --port 3001",
+    "dev:lan": "NODE_OPTIONS='--inspect' next dev --turbo -H 0.0.0.0 --port 3001",
     "format": "pnpm _format --list-different --write",
     "format:check": "pnpm _format --check",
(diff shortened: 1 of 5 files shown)


src/typescript/frontend/package.json
```diff
@@ -108,6 +108,7 @@
     "check:tests": "tsc -p tests/tsconfig.json",
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist && rm -rf .next",
     "dev": "NODE_OPTIONS='--inspect' next dev --turbo --port 3001",
+    "dev:lan": "NODE_OPTIONS='--inspect' next dev --turbo -H 0.0.0.0 --port 3001",
```
xbtmatt, line 111: Nice ty

xbtmatt (commented):
The max-60dvh was for the non-arena hero section- can you make sure the home page with arena disabled doesn't look funky without it? I remember it being there for a reason- but it's possible it was an arbitrary visual thing and not necessary

alnoki (approved):

0xForkh:
The non-arena section height was a bit too much without the max-h. Mainly because of a huge margin.
I reduced the margin and it looks good now on both arena and non-arena section. ce4af73f26aa0febb09d9e7abc9ee5c20607f496

